PR #6, opened 2017-07-17
Enhanced to support notes attribute

Praveengostu:


website/docs/r/compute_vm_instance.html.markdown
```diff
@@ -84,6 +84,7 @@ The following arguments are supported:
 * `private_subnet` - (Optional) Private subnet for the private network interface of the instance. Accepted values are primary private networks and can be found in the  [subnets doc](https://control.softlayer.com/network/subnets).
 * `disks` - (Optional, array) Numeric disk sizes in GBs. Block device and disk image settings for the computing instance. Defaults to the smallest available capacity for the primary disk are used. If an image template is specified, the disk capacity is provided by the template.
 * `user_metadata` - (Optional) Arbitrary data to be made available to the computing instance.
+*   `notes` - (Optional) A note of up to 1000 characters about the VM instance.
```
Praveengostu: Add validate func only for VM instance or for all other resources supporting notes

ashishth09: Wherever it is defined in the API. There is limitation on the storage notes as well. For example upto 400 characters but I don't see it yet in the public API. So just do for VM alone.

Praveengostu: Added the validate func for notes and added a UAT to verify it
--- PASS: TestAccIBMComputeVmInstance_InvalidNotes (0.01s)
PASS

Issue #18, opened 2017-08-26
Enhancement: update softlayer_lb_vpx resource to have management_ip_address attribute 


tomwganem:
related issue:
https://github.com/softlayer/terraform-provider-softlayer/issues/167

When configuring vpx services to report the source ip of connections to the server, one needs to configure the default gateway of the server to the management ip address of the vpx.

ashishth09:
Closed in #19

PR #19, opened 2017-08-26
set management_ip_address attribute on softlayer_lb_vpx resource

tomwganem:
fix for #18

### DIFF
--- a/ibm/resource_ibm_lb_vpx.go
+++ b/ibm/resource_ibm_lb_vpx.go
@@ -137,6 +137,11 @@ func resourceIBMLbVpx() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+
+			"management_ip_address": {
+				Type:     schema.TypeString,
+				Computed: true,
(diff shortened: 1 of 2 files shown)


ibm/resource_ibm_lb_vpx.go
```diff
@@ -528,6 +533,7 @@ func resourceIBMLbVpxRead(d *schema.ResourceData, meta interface{}) error {
 
 	d.Set("vip_pool", vips)
 	d.Set("ip_count", ipCount)
+	d.Set("management_ip_address", *getObjectResult.ManagementIpAddress)
```
ashishth09: I have found that `Set` API takes care of the pointer indirection. Though at places you would still see the explicit way of doing as you have done before, but in newer codes I think it makes sense to leverage that. So you could just say `d.Set("management_ip_address", getObjectResult.ManagementIpAddress)`. This has the added benefit that if pointer is nil it won't blow even if you have not checked the `nilness` as the `Set` takes care of that.

tomwganem:
@ashishth09 Done.

PR #29, opened 2017-09-20
security group functions

suppandi:


ashishth09:
Hi @suppandi ,

Thanks for submitting this PR 👍 . The offering is still in beta but we can definitely take a look into what we have here. Great effort.

We can work towards fixing some minor problems which I see here and take it closer to the merge-able state. Upfront two major things that we will require before we merge this are:
* Documentation updates
* UAT (User acceptance tests) for each resource changed/created, please follow the readme on how to set up UAT.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return resourceIBMSecurityGroupRuleRead(d, meta)
+}
+
+func resourceIBMSecurityGroupRuleRead(d *schema.ResourceData, meta interface{}) error {
+	sess := meta.(ClientSession).SoftLayerSession()
+	service := services.GetNetworkSecurityGroupService(sess)
+
+	matchingrule, err := findMatchingRule(d, service)
```
ashishth09: We don't have API to read back a rule based on its id? If so we should use that.
If not then, what happens if the id matches but other fields are different. In that case we shouldn't throw error that rule doesn't exist but we should read back all other attributes of that rule so that `terraform` can correctly show what has changed if at all in the form of a diff on next plan or so. Possibly since the fields are not updatable this might not happen. We need to rethink this.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,196 @@
...
+var ALLOWED_DIRECTIONS = []string{"ingress", "egress"}
+var ALLOWED_ETHERTYPES = []string{"IPv4", "IPv6"}
+var ALLOWED_PROTOCOLS = []string{"icmp", "tcp", "udp"}
+
+func resourceIBMSecurityGroupRule() *schema.Resource {
+	return &schema.Resource{
+		Create:   resourceIBMSecurityGroupRuleCreate,
+		Read:     resourceIBMSecurityGroupRuleRead,
+		Delete:   resourceIBMSecurityGroupRuleDelete,
+		Importer: &schema.ResourceImporter{},
```
ashishth09: Can we add an exists method here?

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,196 @@
...
+				Required: true,
+				ForceNew: true,
+			},
+			"protocol": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "icmp, tcp or udp",
+				ForceNew:    true,
+			},
+			"securitygroupid": {
```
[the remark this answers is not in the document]
sakshiag: In the security group rule , we have remote group id and remote ip as a part of schema. Can you add these two  parameters also?

bonifaido:
Any updates on this one?

suppandi:
still making changes requested.

suppandi:
Please review changes.

ibm/resource_ibm_securitygroup_rule.go
ashishth09: Since we already have the rule id here can't we leverage that?

```go

	rules, err := service.
		Mask("id,portRangeMax").
		Filter(
			filter.Build(
				filter.Path("rules.id").Eq(d.Id()),
			),
		).Id(id).GetRules()
```
This will give all the info which we should setback.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,171 @@
...
+		return err
+	}
+
+	if matchingrule == nil {
+		return fmt.Errorf("Could not retrieve matching Security Group Rule")
+	}
+
+	d.SetId(strconv.Itoa(*matchingrule.Id))
+
+	d.Set("id", matchingrule.Id)
```
ashishth09: Not required. `d.SetId(strconv.Itoa(*matchingrule.Id))` should be used to set the implicit resource ID.

ashishth09: Please also add an exist method on the schema like `resourceIBMSecurityGroupExists` or handle the non-existence of the rule here itself in a similar way as done in `resourceIBMSecurityGroupRead`

```go
		// If the group is somehow already destroyed, mark as
		// succesfully gone
		if err, ok := err.(sl.Error); ok && err.StatusCode == 404 {
			d.SetId("")
			return nil
		}
```
You should check if group exists and the rule exists

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,171 @@
...
+	return nil
+}
+
+func resourceIBMSecurityGroupRuleDelete(d *schema.ResourceData, meta interface{}) error {
+	sess := meta.(ClientSession).SoftLayerSession()
+	service := services.GetNetworkSecurityGroupService(sess)
+
+	matchingrule, err := findMatchingRule(d, service)
```
ashishth09: I don't think you need this. It can be considered as an optimization to see if the rule already exists before deleting.  ` _, err = service.Id(*matchingrule.SecurityGroupId).RemoveRules(ids_to_delete)` should suffice.

ibm/resource_ibm_compute_vm_instance.go
```diff
@@ -171,6 +179,14 @@ func resourceIBMComputeVmInstance() *schema.Resource {
 				Computed: true,
 			},
 
+			"private_security_groups": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+				ForceNew: false,
```
ashishth09: Did you mean true? Default is anyway false.

ibm/resource_ibm_compute_vm_instance.go
```diff
@@ -495,18 +514,41 @@ func getVirtualGuestTemplateFromResourceData(d *schema.ResourceData, meta interf
 			return opts, fmt.Errorf("Error creating virtual guest: %s", err)
 		}
 		primaryNetworkComponent.NetworkVlan.PrimarySubnetId = &primarySubnetID
+		usePrimaryNetworkComponent = true
+	}
+
+	// Apply security groups if provided
+	publicSecurityGroups := d.Get("public_security_groups")
+	if publicSecurityGroups != nil {
```
ashishth09: I don't think it will ever be nil.

ibm/resource_ibm_compute_vm_instance.go
```diff
@@ -495,18 +514,41 @@ func getVirtualGuestTemplateFromResourceData(d *schema.ResourceData, meta interf
...
+			len(publicSecurityGroupsList))
+		for i, v := range publicSecurityGroupsList {
+			sgid, err := findSecurityGroupByName(v.(string), meta)
+			if err != nil {
+				return opts, fmt.Errorf("Error finding security group: %s", err)
+			}
+			publicSGBindings[i] = datatypes.Virtual_Network_SecurityGroup_NetworkComponentBinding{
+				SecurityGroupId: sl.Int(sgid),
+			}
+		}
```
ashishth09: Since both `public_security_groups` and` private_security_groups` field have a common way of composing the 
SecurityGroupBindings , could you pull the logic in a function and use it across.

Typically such functions are expanders and you can call them 
expandSecurityGroupBindings or something like that. This function could return 
`[]datatypes.Virtual_Network_SecurityGroup_NetworkComponentBinding`
 

Note: Not needed as group names should be dropped and ids be used instead.

ibm/resource_ibm_compute_vm_instance.go
```diff
@@ -157,6 +157,14 @@ func resourceIBMComputeVmInstance() *schema.Resource {
 				Computed: true,
 			},
 
+			"public_security_groups": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
```
ashishth09: Can we have this exported as `public_security_group_ids`? Same with the private counterpart. 
Typically the security groups will be created using terraform itself and then its id can be easily interpolated and used here. It also matches with what the SL API expects and this case it is a good candidate to keep it that way.

In cases security groups are created outside terraform we can provide a datasource and use that. There is also a mechanism of importing existing ones.

ibm/resource_ibm_securitygroup.go
```diff
@@ -0,0 +1,185 @@
...
+
+	name := d.Get("name").(string)
+	description := d.Get("description").(string)
+
+	groups, err := services.GetAccountService(sess).
+		Filter(filter.Path("securityGroups.name").Eq(name).Build()).
+		GetSecurityGroups()
+
+	if nil == err && len(groups) > 0 {
+		group := groups[0]
```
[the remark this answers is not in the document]
suppandi: Yes. names need not be unique. But we need to get a way to use the predefined security groups like allow_https and allow_all. We cant make them remember ids. This lets them define an existing group by name and use it in a vm.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,171 @@
...
+
+	for _, rule := range rules {
+		if direction == *rule.Direction &&
+			ethertype == *rule.Ethertype &&
+			portrangemin == *rule.PortRangeMin &&
+			portrangemax == *rule.PortRangeMax &&
+			protocol == *rule.Protocol {
+			return &rule, nil
+		}
+	}
```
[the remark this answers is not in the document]
suppandi: will look into this

ibm/resource_ibm_securitygroup.go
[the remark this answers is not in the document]
ashishth09: Sure, we don't want folks remember the ids, for that we are going to provide a data source to get the required id.

Writing a Data source mostly entails picking code from your read method. I can cover that in parallel if you want as I don't want to slow you down.

ibm/resource_ibm_compute_vm_instance.go
ashishth09: https://github.com/IBM-Bluemix/terraform-provider-ibm/pull/29/commits/468b4e658c5a2cbd823c9de166e745fb45dcacd0#diff-e4b220657319756663dd781aea77b4d0R446

will be called for each VM to translate the name to id, not an efficient way if someone spins 100 of VMs or so. 

It should be an ID. We will add a DS which gets us the ID of the group by taking name as input and use it anywhere, any number of times we like.

suppandi:
hopefully made all the changes. Still missing UAT. will work on them next week.

rahuldivgan:
Any idea on expediting this? This feature would really help. Thanks.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,250 @@
...
+		return fmt.Errorf("Error retrieving Security Group Rule: %s", err)
+	}
+
+	if len(matchingrules) == 0 {
+		d.SetId("")
+		return nil
+	}
+
+	if matchingrules[0].Direction != nil {
+		d.Set("direction", matchingrules[0].Direction)
```
ashishth09: There is no need to check for the `nilness` as d.Set takes care of that. You could remove that. Same goes for the below code.
As you have already noted in that case there is no need to deference the pointer also.

vendor/github.com/softlayer/softlayer-go/services/network.go
```diff
@@ -6301,7 +6301,7 @@ func (r Network_SecurityGroup) Offset(offset int) Network_SecurityGroup {
 }
 
 // Add new rules to a security group by sending in an array of template [[SoftLayer_Network_SecurityGroup_Rule (type)]] objects to be created.
-func (r Network_SecurityGroup) AddRules(ruleTemplates []datatypes.Network_SecurityGroup_Rule) (resp bool, err error) {
+func (r Network_SecurityGroup) AddRules(ruleTemplates []datatypes.Network_SecurityGroup_Rule) (resp string, err error) {
```
ashishth09: With the latest API I see that the return result is `SoftLayer_Network_SecurityGroup_Request` per
http://sldn.softlayer.com/reference/services/softlayer_network_securitygroup/addrules.

Please test by vendoring latest softalyer-go.  You can generate the latest softayer-go by issuing `make generate` in the softlayer-go project. We should try with that updated library. I will raise issue in softlayer-go for this once we can confirm the latest API so obtained works fine. This should be used to validate the UAT.


```go

// Add new rules to a security group by sending in an array of template [[SoftLayer_Network_SecurityGroup_Rule (type)]] objects to be created.
func (r Network_SecurityGroup) AddRules(ruleTemplates []datatypes.Network_SecurityGroup_Rule) (resp datatypes.Network_SecurityGroup_Request, err error) {
	params := []interface{}{
		ruleTemplates,
	}
	err = r.Session.DoRequest("SoftLayer_Network_SecurityGroup", "addRules", params, &r.Options, &resp)
	return
}
```

ibm/resource_ibm_securitygroup.go
```diff
@@ -0,0 +1,195 @@
...
+		Update:   resourceIBMSecurityGroupUpdate,
+		Delete:   resourceIBMSecurityGroupDelete,
+		Exists:   resourceIBMSecurityGroupExists,
+		Importer: &schema.ResourceImporter{},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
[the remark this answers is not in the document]
suppandi: fixed

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/softlayer/softlayer-go/sl"
+)
+
+func resourceIBMSecurityGroupRule() *schema.Resource {
+	return &schema.Resource{
+		Create:   resourceIBMSecurityGroupRuleCreate,
+		Read:     resourceIBMSecurityGroupRuleRead,
+		Delete:   resourceIBMSecurityGroupRuleDelete,
+		Exists:   resourceIBMSecurityGroupRuleExists,
+		Importer: &schema.ResourceImporter{},
```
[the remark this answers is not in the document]
suppandi: yes. but it doesnt make any functional difference between updating in place or deleting and adding back. I think we should leave it as is.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,250 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"direction": {
+				Type:         schema.TypeString,
```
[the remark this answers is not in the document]
suppandi: fixed.

vendor/github.com/softlayer/softlayer-go/services/network.go
suppandi: I think the api is going to change more. Right now it returns a request object that looks for example like `{"id":"98bb9c91fcdad1fa29d9a9d"}`. That id field does not seem to match any api to reuse. So I let it go as a string for now. We can change it later when it is more useful.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,253 @@
...
+				Optional:     true,
+				Description:  "icmp, tcp or udp",
+				ForceNew:     true,
+				ValidateFunc: validateSecurityRuleProtocol,
+			},
+			"security_group_id": {
+				Type:     schema.TypeInt,
+				Required: true,
+				ForceNew: true,
+			},
```
rlrossiter: You're missing the field remote_ip (http://sldn.softlayer.com/reference/datatypes/SoftLayer_Network_SecurityGroup_Rule)
For verification, it can either be a CIDR or a single IP address.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,253 @@
...
+	// if only one of min/max is provided, set the other one to the provided
+	if sgrule.PortRangeMin != nil && sgrule.PortRangeMax == nil {
+		sgrule.PortRangeMax = sgrule.PortRangeMin
+	}
+	if sgrule.PortRangeMax != nil && sgrule.PortRangeMin == nil {
+		sgrule.PortRangeMin = sgrule.PortRangeMax
+	}
+
+	if sgrule.PortRangeMin != nil {
+		if *sgrule.PortRangeMax < *sgrule.PortRangeMin {
```
rlrossiter: FYI, portRangeMin and portRangeMax when used with the protocol ICMP are reused as the ICMP type and code respectively. So allowing ICMP ping (type 8, code 0) would be protocol ICMP, portRangeMin 8, portRangeMax 0

We do have appropriate error checking in the API for this, so do you want to lean on that error checking instead of reimplementing it here?

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	filters = append(filters, filter.Path("rules.direction").Eq(rule.Direction))
+
+	if rule.Ethertype != nil {
+		filters = append(filters, filter.Path("rules.ethertype").Eq(rule.Ethertype))
+	}
+	if rule.Protocol != nil {
+		filters = append(filters, filter.Path("rules.protocol").Eq(rule.Protocol))
+	}
```
rlrossiter: These filters as they currently stand may call 2 rules identical that are not equivalent.

You may have a rule with port 22 open for TCP to 192.168.0.0/24, and another rule for port 22 TCP to remoteGroup 129347. These aren't the same but would be called so with this function.

You'll need to add filters for remoteIp and remoteGroup as well here.

ibm/resource_ibm_securitygroup_rule.go
```diff
@@ -0,0 +1,253 @@
...
+	return nil
+}
+
+func resourceIBMSecurityGroupRuleDelete(d *schema.ResourceData, meta interface{}) error {
+	sess := meta.(ClientSession).SoftLayerSession()
+	service := services.GetNetworkSecurityGroupService(sess)
+
+	id, _ := strconv.Atoi(d.Id())
+	ids_to_delete := []int{id}
+	_, err := service.Id(d.Get("security_group_id").(int)).RemoveRules(ids_to_delete)
```
rlrossiter: If you want to ensure that a security group is deleted in this function, along with removing all rules, you'll also want to detach all attached interfaces from it. A security group with any attached interfaces cannot be deleted.

It can also be more complex, but if this security group is referenced as a remote security group by a rule in any other groups, it also cannot be deleted. The IDs of the groups where they are still referenced will be returned as an error in the delete, so you can use that message to find where to clean up more of these rules.

website/docs/r/compute_vm_instance.html.markdown
```diff
@@ -33,6 +33,7 @@ resource "ibm_compute_vm_instance" "twc_terraform_sample" {
     local_disk = false
     public_vlan_id = 1391277
     private_vlan_id = 7721931
+    public_security_groups = ["allow_ssh"]
```
rlrossiter: it looks like you have `public_security_groups` here with names, but it looks like you have `public_security_group_ids` below with int ids.

website/docs/r/security_group_rule.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+
+## Example Usage
+
+```
+resource "ibm_security_group_rule" "allow_port_8080" {
+    direction = "ingress"
+    ether_type = "IPv4"
+    port_range_min = 8080
+    port_range_max = 8080
+    protocol = tcp
```
rlrossiter: I think this needs to be a string

website/docs/r/security_group_rule.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `direction` - (Optional, string) Direction of traffic: ingress or egress
+* `ether_type` - (Optional, string) IP version: IPv4 or IPv6 (case sensitive). Defaults to 'IPv4'
+* `port_range_min` - (Optional, int) Lower bound of port range to allow
+* `port_range_max` - (Optional, int) Upper bound of port range to allow
+* `protocol` - (Optional, string) Traffic protocol: icmp or tcp or udp (case sensitive)
+* `security_group_id` - (Required, string) Member of this security group
```
rlrossiter: the datatype for this should be an int since it's an ID, correct?

ibm/resource_ibm_compute_vm_instance.go
```diff
@@ -194,6 +204,16 @@ func resourceIBMComputeVmInstance() *schema.Resource {
...
 			},
 
+			"private_security_group_ids": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeInt},
+				Set: func(v interface{}) int {
+					return v.(int)
+				},
+				ForceNew: true,
```
ashishth09: Since user's can  add more groups after it is provisioned. I think it should be an updatable field. Same with `public_security_group_ids`.
Only thing is one will need a soft reboot of the VSI in such cases and should be properly documented.

ashishth09: I think this might be limited then. Probably we can  add another resource  like 
```
resource ibm_security_group_ifc_binding {
   security_group_id = "${...}"
   virtual_guest_id = "${...}"
   interface_type = "PUBLIC"
}
```
But each addition requires a reboot which I think is not good if we were to add 5 such bindings. This could be used to mitigate the above problem and a more common case would be rules are already created and user would specify them during the VSI placeOrder. 
These are just some of my initial thoughts. Definitely needs some more rounds of it.

rlrossiter: Additions only require a reboot if you're adding a security group to a network component that has never had security groups on it before.

You can check this by calling https://sldn.softlayer.com/reference/services/SoftLayer_Virtual_Guest_Network_Component/securityGroupsReady

If there is a security group on the network component, and securityGroupsReady returns False, then a reboot is required. Otherwise, any security group application to that network component will take effect without reboot.

ashishth09: Awesome 👍, will check this.

PR #40, opened 2017-10-04
vendor latest bluemix-go

Praveengostu:


ibm/data_source_ibm_container_cluster.go
```diff
@@ -105,11 +145,35 @@ func dataSourceIBMContainerClusterRead(d *schema.ResourceData, meta interface{})
 		boundedService["namespace"] = service.Namespace
 		boundedServices = append(boundedServices, boundedService)
 	}
-
+	//Set Vlans
+	vlansR := clusterFields.Vlans
+	vlans := make([]map[string]interface{}, 0, len(vlansR))
```
ashishth09: Since we already know the length, using make([]map[string]interface{}, len(vlansR)) and then vlans[i] = ... is better I guess

ashishth09: On second though.. how about pulling this thing in a function and calling it flattenVlans in the strcutures.go

Praveengostu: Done

PR #41, opened 2017-10-04
Rachael graham doc edits

Rachael-Graham:
Doc edits for the new provider push


website/docs/index.html.markdown
```diff
@@ -101,15 +101,15 @@ $ terraform plan
 
 The following arguments are supported in the `provider` block:
 
-* `bluemix_api_key` - (optional) The Bluemix API key. You must either add it as a credential in the provider block or source it from the `BM_API_KEY` (higher precedence) or `BLUEMIX_API_KEY` environment variable. The key is required to provision Cloud Foundry or IBM Container Service resources, such as any resource that begins with `ibm` or `ibm_container`.
```
ashishth09: In the provider block this is an optional parameter. Let us keep it that way.

website/docs/index.html.markdown
```diff
@@ -101,15 +101,15 @@ $ terraform plan
...
 
-* `bluemix_timeout` - (optional) The timeout, expressed in seconds, for interacting with Bluemix APIs. You can also source the timeout from the `BM_TIMEOUT` (higher precedence) or `BLUEMIX_TIMEOUT` environment variable. The default value is `60`.
+* `bluemix_timeout` - (Optional, integer) The timeout, expressed in seconds, for interacting with Bluemix APIs. You can also source the timeout from the `BM_TIMEOUT` (higher precedence) or `BLUEMIX_TIMEOUT` environment variable. The default value is `60`.
 
-* `softlayer_username` - (optional) The SoftLayer user name. You must either add it as a credential in the provider block or source it from the `SL_USERNAME` (higher precedence) or `SOFTLAYER_USERNAME` environment variable.
+* `softlayer_username` - (Required for Softlayer resources, string) The SoftLayer user name. You must either add it as a credential in the provider block or source it from the `SL_USERNAME` (higher precedence) or `SOFTLAYER_USERNAME` environment variable.
```
ashishth09: I think  `Required for Softlayer resources` should be dropped and  `(optional, string)` be used as in the provider block it is still an optional field. In fact all of the parameters are optional in the provider block as they can be given values via environment variables for each of them or they already have a default value. We have correctly called out with the phrase ` You must either` to emphasize it. This also aligns with what other providers have done for example - https://www.terraform.io/docs/providers/azurerm/index.html, https://www.terraform.io/docs/providers/aws/index.html etc. 

To emphasize the requirement for SL resources ,  we can frame it as 

`You must either add it as a credential in the provider block or source it from the `SL_API_KEY` (higher precedence) or `SOFTLAYER_API_KEY` environment variable if you are provisioning any SoftLayer resources`

website/docs/index.html.markdown
```diff
@@ -101,15 +101,15 @@ $ terraform plan
...
 
-* `softlayer_api_key` - (optional) The SoftLayer API key. You must either add it as a credential in the provider block or source it from the `SL_API_KEY` (higher precedence) or `SOFTLAYER_API_KEY` environment variable. The key is required to provision SoftLayer resources, such as any resource that begins with `ibm_compute`.
+* `softlayer_api_key` - (Required for Softlayer resources, string) The SoftLayer API key. You must either add it as a credential in the provider block or source it from the `SL_API_KEY` (higher precedence) or `SOFTLAYER_API_KEY` environment variable. The key is required to provision SoftLayer resources, such as any resource that begins with `ibm_compute`.
```
ashishth09: Same as `softlayer_username`

website/docs/index.html.markdown
Rachael-Graham: These changes were requests from @davetropeano

ashishth09: Sure the reason for this is explained in the below review comment.

davetropeano:
There is a readability problem with something being "required" while at the same time "optional".

Given that we expect Schematics to be the main user of the Cloud Provider the attributes like `softlayer_username` are actually required. I think @Rachael-Graham 's rewording better captures the nature of these attributes.

ashishth09:
Can we not put the strict requirement of these fields in the Schematics documentation? 
Something like Schematics doesn't allow you to export environment fields and hence 
you should use the `Static credentials` way of setting up the provider in Schematics,
as explained in the provider documentation.

ashishth09:
I have picked all the four commits except the change to index.html one. We needed this to verify the documentation once from hosting it without waiting for all to be merged.

website/docs/r/subnet.html.markdown
```diff
@@ -8,30 +8,20 @@ description: |-
...
- Number of usuable IP addresses is `capacity` - 3. If `capacity` is 4, the number of usuable IP addresses is 4 - 3 = 1. If `capacity` is 8, the 
- number of usuable IP addresses is 8 - 3 = 5. For additional details, refer to [Static and Portable IP blocks](https://knowledgelayer.softlayer.com/articles/static-and-portable-ip-blocks).
+addresses for IBM resources in the VLAN. Because each portable subnet has a default gateway IP address, network IP address, and broadcast IP address, the number of usuable IP addresses is `capacity` - 3. A `capacity` of 4 means that the number of usuable IP addresses is 1; a `capacity` of 8 means that the number of usuable IP addresses is 5. For example, consider a portable subnet of `10.0.0.0/30` that has `10.0.0.1` as a default gateway IP address, `10.0.0.0` as a network IP address, and `10.0.0.3` as a broadcast IP address. Only `10.0.0.2` can be assigned to IBM resources as a secondary IP address. For additional details, refer to [Static and Portable IP blocks](https://knowledgelayer.softlayer.com/articles/static-and-portable-ip-blocks).
```
knlyons: Should `usuable IP addresses` be `usable IP addresses`?

ashishth09: yes, nice catch.

website/docs/r/subnet.html.markdown
```diff
@@ -8,30 +8,20 @@ description: |-
...
-a static subnet on the public IP address of the virtual server. Unlike the portable subnet, `capacity` is same with a number of usuable IP address. 
-For example, if a static subnet is `10.0.0.0/30`, `capacity` is 4 and four IP addresses(10.0.0.0 ~ 10.0.0.3) can be used as secondary IP addresses. 
-For additional details, refer to [Subnet](https://knowledgelayer.softlayer.com/topic/subnets).
+The static IPv4 subnet provides secondary IP addresses for primary IP addresses. It provides secondary IP addresses for IBM resources such as virtual servers, bare metal servers, and netscaler VPXs. Consider a virtual server that requires secondary IP addresses. Users can create a static subnet on the public IP address of the virtual server. Unlike the portable subnet, the number of usuable IP addresses for the stactic subnet is the same as the value of `capacity`. For example, when a static subnet of `10.0.0.0/30` has a `capacity` of 4, then four IP addresses (10.0.0.0 - 10.0.0.3) can be used as secondary IP addresses. For additional details, refer to [Subnet](https://knowledgelayer.softlayer.com/topic/subnets).
```
knlyons: `usuable IP addresses` should also be fixed here to say `usable IP addresses`

knlyons:
@Rachael-Graham I found some typos in a couple of files, but otherwise LGTM!

PR #43, opened 2017-10-05
Updaed cluster Datasource doc with Vlans

Praveengostu:


website/docs/d/container_cluster.html.markdown
```diff
@@ -41,3 +41,11 @@ The following attributes are exported:
 * `worker_count` - The number of workers that are attached to the cluster.
 * `workers` - The IDs of the workers that are attached to the cluster.
 * `bounded_services` - The services that are bounded to the cluster.
+* `VLANS` - The VLAN'S that are attached to the cluster. Nested `VLANS` blocks have the following structure:
```
ashishth09: I think you meant vlans as the attribute name.

Praveengostu: Yes. Done with changes

PR #74, opened 2017-11-10
Network Gateway resource support files

favrecr:
Hello folks, please find in this PR the code for supporting ordering of a Vyatta Network Gateway, changes to the following files:
1- Added resource_ibm_network_gateway.go
2- Modified provider.go to support the added resource

Feel free to reach me for changes /  feedback


ashishth09:
Hi @favrecr,

 Thanks for the PR 👍.  
A user can specify all the details mentioned here - https://github.com/IBM-Bluemix/terraform-provider-ibm/pull/74/files#diff-40d0ac9f649b84bca138b4f7998b16beR611 and be able to provision the gateway without with the `ibm_bare_metal` resource. Is that a possibility?

favrecr:
Hello @ashishth09, certainly I tried that way first but ibm_bare_metal has hard coded the filter to look for package BARE_METAL_CPU and the gateway appliances are NETWORK_GATEWAY_APPLIANCE, so in this case the filter is applied differently, also an enhancement that I did was to setup a flag for order verification, so in case of Vyatta Network Gateways the order is verified before actually putting it. Please let me know your thoughts.

favrecr:
@ashishth09 also I would like to mention that orders for Vyatta Network gateways must be submitted as child containers, so the structure is different than a regular bare metal order. Please check this post that shows an example
https://stackoverflow.com/questions/36475885/order-vyatta-and-netscaler-vpx-with-java

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,996 @@
...
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"private_network_only": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: true,
+			},
+			"debug": {
```
ashishth09: I think we don't really need this. You can go ahead and do the verification of the order  before the actual order in create method regardless of debug. Typically `terraform` users will resort to `terraform plan` to see if all is OK and since debug is part of the create method, it is not going to hit this path. So assume debug is on for other paths.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,996 @@
...
+	}
+
+	billingItem, err := service.Id(id).GetBillingItem()
+	if err != nil {
+		return fmt.Errorf("Error getting billing item for Network Gateway: %s", err)
+	}
+
+	// Monthly  Softlayer items only support an anniversary date cancellation option.
+	billingItemService := services.GetBillingItemService(sess)
+	_, err = billingItemService.Id(*billingItem.Id).CancelItem(
```
ashishth09: Is it possible for billingItem to be nil if some one cancelled the device outside terraform? In that case `*billingItem.Id` will blow. I see a similar problem in `ibm_bare_metal`. We should validate this and check for nilness if that is the case.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,996 @@
...
+			"unbonded_network": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: true,
+			},
+
+			// Monthly only
+			"public_bandwidth": {
+				Type:             schema.TypeString,
```
ashishth09: Should it have a default value of 20000? I see later you have it hard-coded as such.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,996 @@
...
+	return result.Id != nil && *result.Id == id, nil
+}
+
+func getMonthlyGatewayOrder(d *schema.ResourceData, meta interface{}) (datatypes.Container_Product_Order, error) {
+
+	d.Set("package_key_name", "NETWORK_GATEWAY_APPLIANCE")
+	d.Set("os_reference_code", "OS_VYATTA_5600_5_X_UP_TO_1GBPS_SUBSCRIPTION_EDITION_64_BIT")
+	d.Set("process_key_name", "INTEL_SINGLE_XEON_1270_3_40_2")
+	d.Set("os_key_name", "OS_VYATTA_5600_5_X_UP_TO_1GBPS_SUBSCRIPTION_EDITION_64_BIT")
+	d.Set("public_bandwidth", "BANDWIDTH_20000_GB")
```
ashishth09: Shouldn't thes come from user input rather than being hard-coded. As package_key_name is always going to have `NETWORK_GATEWAY_APPLIANCE` as value can we we drop
```
+			// Monthly only
+			"package_key_name": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: applyOnce,
+			},
```

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,996 @@
...
+			availableModels = availableModels + ", "
+		}
+		if *pkg.KeyName == model {
+			return pkg, nil
+		}
+	}
+	return datatypes.Product_Package{}, fmt.Errorf("No Gateway package key name for %s. Available package key name(s) is(are) %s", model, availableModels)
+}
+
+func getItemPriceIdGateway(items []datatypes.Product_Item, categoryCode string, keyName string) (datatypes.Product_Item_Price, error) {
```
ashishth09: Can't we re-use the `getItemPriceId` from `ibm_bare_metal` implementation?

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,996 @@
...
+		Timeout:        24 * time.Hour,
+		Delay:          10 * time.Second,
+		MinTimeout:     1 * time.Minute,
+		NotFoundChecks: 24 * 60,
+	}
+
+	return stateConf.WaitForState()
+}
+
+func findMemoryItemPriceIdGateway(items []datatypes.Product_Item, d *schema.ResourceData) (datatypes.Product_Item_Price, error) {
```
ashishth09: Can't we re-use findMemoryItemPriceId from `ibm_bare_metal` implementation?

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,996 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return datatypes.Product_Item_Price{},
+		fmt.Errorf("Could not find the price item for %d GB memory. Available items are %s", memory, availableMemories)
+}
+func findNetworkItemPriceIdGateway(items []datatypes.Product_Item, d *schema.ResourceData) (datatypes.Product_Item_Price, error) {
```
ashishth09: Can't we re-use `findNetworkItemPriceId` from `ibm_bare_metal` implementation?

favrecr:
@ashishth09 thanks for the comments, I will address those and update this PR.

ibm/resource_ibm_network_gateway.go
favrecr: @ashishth09 these are the values supported for the configuration:
BANDWIDTH_0_GB ( 0 GB Bandwidth ) , 
BANDWIDTH_UNLIMITED_100_MBPS_UPLINK ( Unlimited Bandwidth (100 Mbps Uplink) ) , BANDWIDTH_0_GB_2 ( 0 GB Bandwidth ) , 
BANDWIDTH_20000_GB ( 20000 GB Bandwidth ) , 

I added default to be the most common one BANDWIDTH_20000_GB

ibm/resource_ibm_network_gateway.go
favrecr: @ashishth09 I removed the logic for the DEBUG, included the VerifyOrder and adjusted the error message showed if it fails

ibm/resource_ibm_network_gateway.go
favrecr: I added this code:
	billingItem, err := service.Id(id).GetBillingItem()
	if err != nil || *billingItem.Id == 0 {
		return fmt.Errorf("Error getting billing item for Network Gateway: %s", err)
	}
Where assuming that billingItem.Id would get a zero value in case that the item has been deleted

ibm/resource_ibm_network_gateway.go
favrecr: Included all values in Default properties, also will document them as variables for variables.tf files

ibm/resource_ibm_network_gateway.go
favrecr: All instances of getItemPriceIdGateway have been changed to use getItemPriceId

ibm/resource_ibm_network_gateway.go
favrecr: All instances of findMemoryItemPriceIdGateway have been changed to findMemoryItemPriceId

ibm/resource_ibm_network_gateway.go
favrecr: All instances of findNetworkItemPriceIdGateway have been changed to findNetworkItemPriceId

favrecr:
@ashishth09 Please review this PR, all changes have been completed and the basic UAT has been included, it was also ran and provisioned the network gateway as expected.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,896 @@
...
+				Default:  false,
+				ForceNew: true,
+			},
+
+			// Monthly only
+			"public_bandwidth": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				ForceNew:         true,
+				Default:          "BANDWIDTH_20000_GB",
```
ashishth09: Won't it make more sense to have it just 20000 and you compose the actual value in the code?
Also are we sure that this default works in all the cases? You can refer to `public_bandwidth` attribute usage in `ibm_bare_metal` resource.

ibm/resource_ibm_network_gateway_test.go
```diff
@@ -0,0 +1,138 @@
...
+			hostname               = "%s"
+			domain                 = "terraformuat.ibm.com"
+			os_reference_code      = "OS_VYATTA_5600_5_X_UP_TO_1GBPS_SUBSCRIPTION_EDITION_64_BIT"
+			datacenter             = "ams01"
+			network_speed          = 100
+			hourly_billing         = true
+			private_network_only   = false
+			user_metadata          = "No metadata"
+			tcp_monitoring         = true
+			package_key_name       = "NETWORK_GATEWAY_APPLIANCE"
```
ashishth09: The package name should be constant hard-coded value and doesn't need to be configurable by the user

ibm/resource_ibm_network_gateway_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	//"github.com/hashicorp/terraform/helper/schema"
+	"github.com/softlayer/softlayer-go/datatypes"
+	//"github.com/softlayer/softlayer-go/filter"
+	//"github.com/softlayer/softlayer-go/helpers/location"
+	//"github.com/softlayer/softlayer-go/helpers/product"
```
ashishth09: Probably a nitpick, let us remove all these unwanted packages altogether. I am not if you are already using some good IDE which will take care of it automatically when you save the file.

ibm/resource_ibm_network_gateway_test.go
```diff
@@ -0,0 +1,138 @@
...
+	var networkGateway datatypes.Hardware
+	//could also be datatypes.hardware
+
+	hostname := acctest.RandString(16)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
```
ashishth09: How about a checking for the appliance destroy? You can check `testAccCheckIBMComputeBareMetalDestroy` in bare metal test

ibm/resource_ibm_network_gateway.go
ashishth09: I think the check should be more for `billingItem` to be not nil then `*billingItem.Id == 0`.  The latter would still blow if id is `nil` so a check for billingItem.Id != nil is more appropriate I guess.  How about trying it by cancelling the device to see how it behaves.

ibm/resource_ibm_network_gateway_test.go
favrecr: @ashishth09 thanks for all the feedback and comments, I have modified the code to include all the changes.

Please share your thoughts on it.

ibm/resource_ibm_network_gateway_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+	})
+}
+
+func testAccCheckIBMNetworkGatewayConfig_basic(hostname string) string {
+	return fmt.Sprintf(`
+resource "ibm_network_gateway" "terraform-acceptance-test-1" {
+			hostname               = "%s"
+			domain                 = "terraformuat.ibm.com"
+			os_reference_code      = "OS_VYATTA_5600_5_X_UP_TO_1GBPS_SUBSCRIPTION_EDITION_64_BIT"
```
ashishth09: `os_reference_code` is not needed in the schema. `os_key_name` will do the job.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,895 @@
...
+			},
+
+			"network_speed": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  100,
+				ForceNew: true,
+			},
+			//Does not apply for Network gateway but is required for cancellation
+			"hourly_billing": {
```
ashishth09: Why is it required during cancellation 🤔 ? Can't we simply assume it is monthly and use its value as false?

favrecr: This has been changed to reflect a default value of FALSE

ibm/resource_ibm_network_gateway_test.go
favrecr: os_reference_code has been removed

favrecr:
@ashishth09 Please review the latest version, including the documentation page.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+			notification,
+			response,
+			vulnerabilityScanner,
+		},
+	}
+
+	// Add optional price ids.
+	// Add public bandwidth
+
+	if publicBandwidth, ok := d.GetOk("public_bandwidth"); ok {
```
ashishth09: `public_bandwidth` has a default value, hence it will always get one even if user doesn't specify one.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+		if err != nil {
+			return datatypes.Container_Product_Order{}, err
+		}
+		order.Prices = append(order.Prices, bandwidth)
+	}
+
+	// Add prices of disks.
+	var arrayDrives []interface{}
+	var arrayDrivestemp []string
+	arrayDrivestemp = append(arrayDrivestemp, "HARD_DRIVE_1_00_TB_SATA_III")
```
ashishth09: Should this be exposed in schema as `disk_names` or something of that sort like in `ibm_bare_metal`

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+			if err != nil {
+				return datatypes.Container_Product_Order{}, err
+			}
+			order.Prices = append(order.Prices, diskPrice)
+		}
+	}
+
+	// Add storage_groups for RAID configuration
+	//hard coded support for disk controller, just RAID 1 is supported in this release of the resource
+	diskController, err := getItemPriceId(items, "disk_controller", "DISK_CONTROLLER_RAID_1")
```
ashishth09: Should this be exposed in schema and be configurable?

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+	d.SetId(fmt.Sprintf("%d", id))
+
+	// Set tags
+	err = setHardwareTags(id, d, meta)
+	if err != nil {
+		return err
+	}
+
+	var storageIds []int
+	if storageIdsSet := d.Get("file_storage_ids").(*schema.Set); len(storageIdsSet.List()) > 0 {
```
ashishth09: `file_storage_ids` and `block_storage_ids` are not defined on the schema.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+	return &schema.Resource{
+		Create:   resourceIBMNetworkGatewayCreate,
+		Read:     resourceIBMNetworkGatewayRead,
+		Update:   resourceIBMNetworkGatewayUpdate,
+		Delete:   resourceIBMNetworkGatewayDelete,
+		Exists:   resourceIBMNetworkGatewayExists,
+		Importer: &schema.ResourceImporter{},
+
+		Schema: map[string]*schema.Schema{
+			"id": {
```
ashishth09: We will remove `id` from all the schema as newer terraform plugin doesn't allow it to be explicity stated.
Let us remove it. You can continue using d.SetId though.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+				Required: true,
+				ForceNew: true,
+			},
+
+			"notes": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"os_reference_code": {
```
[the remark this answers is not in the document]
favrecr: Fixed, it was a remainder from the manual process of ordering the Network Gateway

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+				ForceNew:         true,
+				DiffSuppressFunc: applyOnce,
+			},
+
+			// Monthly only
+			"process_key_name": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				ForceNew:         true,
+				Default:          "INTEL_SINGLE_XEON_1270_3_40_2",
```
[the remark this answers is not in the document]
favrecr: Documented in network_gateway.html.markdown

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+				DiffSuppressFunc: applyOnce,
+			},
+
+			// Monthly only
+			"os_key_name": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				ForceNew:         true,
+				Default:          "OS_VYATTA_5600_5_X_UP_TO_1GBPS_SUBSCRIPTION_EDITION_64_BIT",
```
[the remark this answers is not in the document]
favrecr: Documented in network_gateway.html.markdown

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+				ForceNew: true,
+			},
+
+			// Monthly only
+			"public_bandwidth": {
+				Type:             schema.TypeInt,
+				Optional:         true,
+				ForceNew:         true,
+				Default:          20000,
+				DiffSuppressFunc: applyOnce,
```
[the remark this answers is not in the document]
favrecr: Documented in network_gateway.html.markdown

ibm/resource_ibm_network_gateway.go
favrecr: Removed and added the following reference to the documentation to avoid confusion:
* `RAID_CONTROLLER` - (reference) this version support a single RAID 1 controller pre configured with two hard drives 1 TB SATA drives

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+		} else {
+			d.Set("unbonded_network", true)
+		}
+	}
+
+	if result.OperatingSystem != nil &&
+		result.OperatingSystem.SoftwareLicense != nil &&
+		result.OperatingSystem.SoftwareLicense.SoftwareDescription != nil &&
+		result.OperatingSystem.SoftwareLicense.SoftwareDescription.ReferenceCode != nil {
+		d.Set("os_reference_code", *result.OperatingSystem.SoftwareLicense.SoftwareDescription.ReferenceCode)
```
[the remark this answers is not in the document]
favrecr: Removed, it was not required.

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+
+	// Validate attributes for network gateway ordering.
+	model := packageKeyName
+
+	datacenter, ok := d.GetOk("datacenter")
+	if !ok {
+		return datatypes.Container_Product_Order{}, fmt.Errorf("The attribute 'datacenter' is not defined.")
+	}
+
+	osKeyName, ok := d.GetOk("os_key_name")
```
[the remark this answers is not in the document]
favrecr: Removed unnecessary GetOK

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,816 @@
...
+	if !ok {
+		return datatypes.Container_Product_Order{}, fmt.Errorf("The attribute 'os_key_name' is not defined.")
+	}
+
+	process_key_name, ok := d.GetOk("process_key_name")
```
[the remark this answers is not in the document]
favrecr: Removed unnecessary GetOK

ibm/resource_ibm_network_gateway.go
favrecr: Removed unnecessary GetOK

ibm/resource_ibm_network_gateway.go
favrecr: Added this note into the documentation: RAID_CONTROLLER - (reference) this version support a single RAID 1 controller pre configured with two hard drives 1 TB SATA drives, the idea is to keep the Network Gateway with minimal internal storage configuration since it’s primary purpose is to serve the network

ibm/resource_ibm_network_gateway.go
favrecr: Added this note into the documentation: RAID_CONTROLLER - (reference) this version support a single RAID 1 controller pre configured with two hard drives 1 TB SATA drives, the idea is to keep the Network Gateway with minimal internal storage configuration since it’s primary purpose is to serve the network

ibm/resource_ibm_network_gateway.go
favrecr: Removed.

favrecr:
@ashishth09 all comments and fixes requested have been completed, please let me know the results of your tests

ZhuangYuZY:
Hi Team, glad to see this pull request, it is what I need also, hope it can release soon. And one more comment, do you support to associate VLan to the Vyatta, then Vyatta can manage the Vlan. There is SL API and UI to do it. I think terraform could also provide this function. Thank you.

ashishth09:
@ZhuangYuZY https://github.com/IBM-Bluemix/terraform-provider-ibm/pull/74/files#diff-5b7a46ee0c55957a6adadb7db49bb91bR28 allows you to add vlans

favrecr:
@ashishth09 @ZhuangYuZY , I am working in https://sldn.softlayer.com/reference/services/SoftLayer_Network_Gateway_Vlan , it is pretty much ready, will finish today and submit for review. Thank you guys.

ZhuangYuZY:
Thank you. It will be good. What is the target day to release this new feature?

ashishth09:
@ZhuangYuZY , the next release of this provider will have it. We are gathering other features which will make part of that release.

ZhuangYuZY:
@ashishth09 do you have a rough day? I want to do my dev plan based on it. Thank you.

vkalangu:
tentatively dec 3rd week

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,849 @@
...
+	if err != nil {
+		return fmt.Errorf(
+			"Error waiting for Gateway (%s) to become ready: %s", d.Id(), err)
+	}
+
+	id := *bm.(datatypes.Hardware).Id
+	d.SetId(fmt.Sprintf("%d", id))
+
+	if v, ok := d.GetOk("associated_vlans"); ok && v.(*schema.Set).Len() > 0 {
+		debugvar := expandVlans(v.(*schema.Set).List())
```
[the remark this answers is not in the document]
favrecr: Change has been implemented

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,849 @@
...
+		Timeout:        24 * time.Hour,
+		Delay:          10 * time.Second,
+		MinTimeout:     1 * time.Minute,
+		NotFoundChecks: 24 * 60,
+	}
+
+	return stateConf.WaitForState()
+}
+
+func flattenVLANInstances(list []datatypes.Network_Gateway_Vlan) []map[string]interface{} {
```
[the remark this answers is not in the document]
favrecr: This was not need, because the flattenVLANInstances function was intended to read the VLANS already configured in the Network Gateway, but that information is not relevant in this case. I removed it from the code

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,836 @@
...
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"private_network_only": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: true,
+			},
+			"associated_vlans": {
```
ashishth09: Can we read back this info and set in the read method? This can be useful while importing an already existing resource in SoftLayer and bringing it under terraform management. Since this is a unique feature and widely usable can we also add a test for this? Is this field not updatable? I don't see this being used in the update method. If not we should set to ForceNew on this?

favrecr: @ashishth09 working in those observations, also I will be adding the code to run the config portion of the resource (Vyatta configuration later today), thank you

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,857 @@
...
+	for i := 0; i < len(processingarray); i++ {
+		bypass := *processingarray[i].BypassFlag
+		networkGatewayID := id
+		networkVlanID := *processingarray[i].NetworkVlanId
+		vlanObjectTemplate := datatypes.Network_Gateway_Vlan{
+			BypassFlag:       &bypass,
+			NetworkGatewayId: &networkGatewayID,
+			NetworkVlanId:    &networkVlanID,
+		}
+		_, err := services.GetNetworkGatewayVlanService(sess).CreateObject(&vlanObjectTemplate)
```
ashishth09: I see there is way to pass multiple objects. Can we we that instead https://sldn.softlayer.com/reference/services/softlayer_network_gateway_vlan/createobjects?

ibm/resource_ibm_network_gateway.go
```diff
@@ -0,0 +1,857 @@
...
+		err := setHardwareNotes(id, d, meta)
+		if err != nil {
+			return err
+		}
+	}
+
+	if d.HasChange("associated_vlans") {
+		if v, ok := d.GetOk("associated_vlans"); ok && v.(*schema.Set).Len() > 0 {
+			associatedVlans := expandVlans(v.(*schema.Set).List())
+			resourceIBMNetworkGatewayVlanAssociate(d, meta, associatedVlans, id)
```
ashishth09: If some vlans are deleted in the config, then those must be deleted via https://sldn.softlayer.com/reference/services/SoftLayer_Network_Gateway_Vlan/deleteObjects.

louiseohalloran:
@ashishth09 @favrecr Can you tell me when it is likely that this new feature will be available? Thank you

ashishth09:
@louiseohalloran ,
Yes the release is planned for this weekend.

ashishth09:
Hi @favrecr ,
We made some design changes to support HA directly from the resource. https://github.com/IBM-Cloud/terraform-provider-ibm/commit/86cbfbc8657d9323cf197e49ebe7b979bebb6dbd.
I am closing this PR. Thank you for working on this.

PR #88, opened 2017-12-04
Allow to set the speed of the private NIC

deniskin82:
Allow to set the speed of the private NIC on a node with private network only NIC.


ashishth09:
closed in 3dfcd43e5eddb4f875a1e90a07f0ec81639ae596

Issue #85, opened 2017-11-20
Thoughts about adding Workers to Container Cluster.


Freyert:
Currently there's no Cluster Worker resource and therefore no way to add Workers to a cluster the terraform way.

If we were to add a Cluster Worker resource, the current API for creating Container Clusters would not mesh well because it makes Clusters dependent on Workers, but a Cluster must be created before a Worker can be created.

The true relation is that a Worker Depends on a Cluster. That is:

```
Worker -> Cluster
and not
Cluster -> Worker
```

I recommend the deprecation of the `ibm_container_cluster.workers` property in favor of creating a resource property `ibm_container_worker.cluster` that indicates the cluster a worker belongs to.

---

Caveats:
* It appears the current implementation only allows you to have a single type of machine resource.
  * The [Cluster API](https://github.com/IBM-Bluemix/bluemix-go/blob/master/api/container/containerv1/clusters.go#L93-L103) seems to dictate the type of worker nodes that can be added to the cluster. _Not_ the [worker API](https://github.com/IBM-Bluemix/bluemix-go/blob/master/api/container/containerv1/workers.go#L62-L66). This eliminates the capability to use a variety of 
nodes in a single cluster.

I think it's still worthwhile to create a Worker resource, but it would be limited in scope to just re-factoring the code within the "cluster resource" that manages workers, to be in an external resource.

This way, in the future the resource would be able to support more complex semantics than are available now; such as many _types_ of workers per cluster.

vkalangu:
@Freyert Thanks....We will look into this soon

ashishth09:
Cluster API has changed now, particularly the worker, we are updating the SDK and then we will have look into it.

Praveengostu:
For the new worker resource, see if the below schema solves the purpose. Review this and let me know your feedback
Schema: map[string]*schema.Schema{
            "cluster": {
                Type:        schema.TypeString,
                Required:    true,
                ForceNew:    true,
                Description: "The cluster name or id",
            },
            "datacenter": {
                Type:     schema.TypeString,
                Required: true,
                ForceNew: true,
            },
            "isolation": {
                Type:     schema.TypeString,
                Required: true,
                ForceNew: true,
            },
            "machine_type": {
                Type:     schema.TypeString,
                Required: true,
                ForceNew: true,
           }
            "privateVlan": {
                Type:     schema.TypeString,
                Required: true,
                ForceNew: true,
            },
            "publicVlan": {
[...]

PR #148, opened 2018-03-06
First draft for adding workers

msvechla:
Some local playground tests I did for adding a worker resource.
Related to: https://github.com/IBM-Cloud/terraform-provider-ibm/issues/85

Still needs a lot of work.


hkantare:
We don't have any separate API to create workers it is deprecated. Workers are created as part of worker_zone.

PR #150, opened 2018-03-12
Autoscale policy

smalpani37:


ashishth09:
Hi @sumit-tembe-ibm ,

This PR deals with more than one issue. I see you have modified vpx , autoscale , cdn and probably more. I am sure they deal with either enhancing or fixing some already existing bugs.
Could you please separate them out and put them in separate PRs for easy review? For example the cdn resource should be one PR, then vpx could be another.
Thanks

PR #166, opened 2018-03-26
Container API changes

Praveengostu:
Make the  Space,org and account guid's are optional part of the new container API changes


vkalangu:
Have we covered all the test aspect of existing cluster (that is a cluster created  with cf ) does not break ? Both co-exist ?

Praveengostu:
Yes we had added UAT for both the cases with and without space,org and account

vkalangu:
MOre specific...you have a single test in e2e with 2 clusters..one with cf and one with out that doing all required k8s activites ?

Praveengostu:
e2e is in a separate repo which will be handled via different PR. Will add specific test with cf and one without it

Praveengostu:
Closing this as to resolve the conflicts.. Will raise a new PR

PR #168, opened 2018-03-29
add support for virtual instance flavor

sakshiag:


sakshiag:
Currently, code contains the logic to create the virtual instance using flavor.
Upgrade is not included.

PR #197, opened 2018-04-20
Add support for datasource resource group

hkantare:


ibm/data_source_ibm_resource_group_test.go
```diff
@@ -0,0 +1,44 @@
...
+				Config: testAccCheckIBMResourceGroupDataSourceConfigWithName(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.ibm_resource_group.testacc_ds_resource_group_name", "name", "default"),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckIBMResourceGroupDataSourceConfigDefault() string {
```
Praveengostu: Both are optional fields can you add a test case if none of them are specified

hkantare: Added testcase to handle is_default false and validating the error.

PR #263, opened 2018-05-30
add computed ip address ids to compute_bare_metal

geogriff-signal:
these ip address ids are needed as input for e.g. ibm_lb_service

also note: i removed "guestNetworkComponentBinding[ipAddressId]" from "primaryVersion6IpAddressRecord" in the object masks in {data,resource}_ibm_compute_bare_metal because they didn't seem to be used.

please let me know if acceptance tests for this are expected, which i didn't write because the existing ip address attributes did not have acceptance tests either.


hkantare:
Thanks for raising PR. Changes looks fine

PR #394, opened 2018-10-11
support NS and SOA DNS record types

Tyrael:
we bumped into the issue that even though that the Softlayer api supports the above mentioned dns record types the provider doesn't have these whitelisted.
I couldn't find any explanation for it except the single line in website/docs/r/dns_record.html.markdown but that sounds bogus as it is perfectly okay to manage a domain in sl/bluemix but delegate a subdomain zone to another name server (for example we have a dns load balancer application which needs the subdomain to be delegated to it).

I've extended the whitelist and the tests and updated the single line of documentation to explain that you don't need to manually set the NS/SOA record for (the top level) domain.


hkantare:
Ok..When I ranUAT it failed 
```
--- FAIL: TestAccIBMDNSRecord_Types (67.28s)
	testing.go:518: Step 0 error: Error applying: 1 error(s) occurred:
		
		* ibm_dns_record.recordSOA: 1 error(s) occurred:
		
		* ibm_dns_record.recordSOA: Error creating DNS Resource soa Record: SoftLayer_Exception_Public: Domain "6vj86k3kdednstest.ibm.com" currently has a SOA record. (HTTP 500)
FAIL
FAIL	github.com/IBM-Cloud/terraform-provider-ibm/ibm	68.812s
```

Can you please delete 'soa' record type in test file, In docs you added an example how to create it that should be enough.

Tyrael:
done

Tyrael:
thank you!

PR #415, opened 2018-11-02
Added support for single vlan FSA firewall

priyanknarvekar:


priyanknarvekar:
@hkantare thanks,
done and updated

PR #446, opened 2018-12-05
Fix erroneous occurence of `machinetype` in docs

brianvans:
Terraform expects `machine_type`:

```
Error: ibm_container_cluster.swiftype-misc: : invalid or unknown key: machinetype
```


jjasghar:
:+1:

Issue #484, opened 2019-01-22
private only container cluster creation hanging


n-marton:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
v0.11.11

### Affected Resource(s)
- ibm_container_cluster

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "ibm_container_cluster" "cluster" {
  name            = "testing"
  datacenter      = "par01"
  disk_encryption = true
  machine_type    = "u2c.2x4"
  hardware        = "shared"
  private_vlan_id = "someid"

  default_pool_size = "1"
}
```

### Expected Behavior
A cluster with only private networking should be created without error.

### Actual Behavior
A cluster is created but the terraform hangs and doesn't exits.

### Steps to Reproduce
1. `terraform apply`

### Cause
https://github.com/IBM-Cloud/terraform-provider-ibm/blob/v0.14.1/ibm/resource_ibm_container_cluster.go#L901 

publicSubnetAdded variable is true even if I create a private network only cluster with no_subnet = false (I want a private subnet attached to my cluster automatically), so WaitForSubnetAvailable function is waiting for an ingress secret which doesn't created if there is no public alb.

### Suggestion
publicSubnetAdded should be false it no_subnet is false and there is no publicVlanID

Praveengostu:
The issue is fixed and will be available in next release

hkantare:
closing the issue

PR #486, opened 2019-01-22
fixing private cluster alb hanging

n-marton:
fixing issue: https://github.com/IBM-Cloud/terraform-provider-ibm/issues/484

### DIFF
--- a/ibm/resource_ibm_container_cluster.go
+++ b/ibm/resource_ibm_container_cluster.go
@@ -900,7 +900,8 @@ func resourceIBMContainerClusterUpdate(d *schema.ResourceData, meta interface{})
 	//TODO put subnet can't deleted in the error message if such case is observed in the chnages
 	var publicSubnetAdded bool
 	noSubnet := d.Get("no_subnet").(bool)
-	if noSubnet == false {
+	publicVlanID := d.Get("public_vlan_id").(string)
+	if noSubnet == false && publicVlanID != "" {
 		publicSubnetAdded = true
 	}
 	if d.HasChange("subnet_id") {


n-marton:
that suggestion won't work in a scenario like this:

```hcl
resource "ibm_container_cluster" "cluster" {
  name            = "testing"
  datacenter      = "par01"
  disk_encryption = true
  machine_type    = "u2c.2x4"
  hardware        = "shared"
  no_subnet = false
  private_vlan_id = "2"
  public_vlan_id = "1"

  default_pool_size = "1"
}
```

If I don't specify neither of the vlans, it must be a free cluster which doesn't have ingress as far as I know.
If I only specify a private vlan, than it have to be a private cluster, which doesn't enable ingress by default, so no ingress domain and ingress secret.
If I specify only a public vlan than it will have an ingress and should wait for it's completion, that goes for the scenario if I specify both like in the example above.

Praveengostu:
If you take a look at documentation here https://console.bluemix.net/docs/containers/cs_cli_reference.html#cs_cli_reference it clearly mentions the private and public vlans are optional and created for the first standard cluster. So we need to handle this case as well

n-marton:
```hcl
resource "ibm_container_cluster" "cluster" {
  name            = "testing"
  datacenter      = "par01"
  disk_encryption = true
  machine_type    = "u2c.2x4"
  hardware        = "shared"

  default_pool_size = "1"
}
```

try to run this

n-marton:
there is some difference between the cli tool's and terraform's behaviour because this tf above doesn't work

anyway if anytime the terraform will work same as the cli in this regard, then it should handle the case you pointed out, I think something like this should do the trick

```go
if (noSubnet == false && publicVlanID != "") ||  (noSubnet == false &&  publicVlanID == "" && privateVlanID == ""){
		publicSubnetAdded = true			
	}		
```

Praveengostu:
@n-marton Sorry for the delayed response.. As mentioned in the doc, just passing the private vlan_id does it restrict creating the public vlan_id for the first standard cluster.. I was thinking if we can fix this by coming up with a `private_only` flag similar to cli.

PR #508, opened 2019-02-08
Change git clone command to http

gcoon151:
Makes instructions more copy and paste friendly.


matthagenbuch:
that is only true if you don't have your ssh key uploaded to github

hkantare:
I think as per above comment you should have your ssh key added

PR #510, opened 2019-02-12
icd base

stevestrutt:
Support for IBM Cloud Databases (ICD). PostgreSQL, Redis, ElasticSearch, Etcd, RabittMQ in all currently supported regions. Mongodb and MySQL are not supported by ICD at this time. 

The following packages need to be vendored in:
bluemix-go - latest level
github.com/hashicorp/terraform/helper/validation


ibm/data_source_ibm_database.go
```diff
@@ -0,0 +1,546 @@
...
+		Read: dataSourceIBMDatabaseInstanceRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "Resource instance name for example, my Database instance",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+
+			"resource_group_id": {
```
hkantare: resource_group_id can be optional...If not provided we shd take default resource grp...the code below already implements..so it can be optional

ibm/data_source_ibm_database.go
```diff
@@ -0,0 +1,546 @@
...
+							Description: "Certificate in base64 encoding",
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"whitelist": {
+				Type:     schema.TypeSet,
+				Optional: true,
```
hkantare: In datasource whitelist is optional or computed..?.From code I see we are fetching from API and set so it can be just computed

ibm/data_source_ibm_database.go
```diff
@@ -0,0 +1,546 @@
...
+			return fmt.Errorf("Error getting user connection string for user (%s): %s", userName, err)
+		}
+		connectionStrings = append(connectionStrings, csEntry)
+	}
+	d.Set("connectionstrings", flattenConnectionStrings(connectionStrings))
+
+	return nil
+}
+
+func dataSourceIBMDatabaseInstanceRead1(d *schema.ResourceData, meta interface{}) error {
```
hkantare: Why do we need tw omethods dataSourceIBMDatabaseInstanceRead and dataSourceIBMDatabaseInstanceRead1?

ibm/resource_ibm_database.go
```diff
@@ -0,0 +1,1086 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"adminuser": {
+				Description: "The admin user id for the instance",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"adminpassword": {
+				Description:  "The admin user id for the instance",
```
hkantare: Description shd be "The admin user passwd..."...If user doesn't provide password will the API set back any password?

ibm/resource_ibm_database.go
```diff
@@ -0,0 +1,1086 @@
...
+	// Original flow had SetId here as its required as input to waitForDatabaseInstanceCreate
+
+	_, err = waitForDatabaseInstanceCreate(d, meta, instance.ID)
+	if err != nil {
+		return fmt.Errorf(
+			"Error waiting for create database instance (%s) to complete: %s", d.Id(), err)
+	}
+
+	icdId := EscapeUrlParm(instance.ID)
+	adminPassword := d.Get("adminpassword").(string)
```
hkantare: we are reading d.Get("adminpassowrd)...in 450 & 456 line

ibm/resource_ibm_database.go
```diff
@@ -0,0 +1,1086 @@
...
+			return fmt.Errorf("Error getting user connection string for user (%s): %s", userName, err)
+		}
+		connectionStrings = append(connectionStrings, csEntry)
+	}
+	d.Set("connectionstrings", flattenConnectionStrings(connectionStrings))
+
+	return nil
+}
+
+func resourceIBMDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{}) error {
```
hkantare: "name & "plan" cannot be updated for a db resource instance?

ibm/data_source_ibm_database.go
stevestrutt: Agreed. These are all optional. Likely cut and paste copy from the resource definition.

ibm/data_source_ibm_database.go
stevestrutt: Omitted to remove. Deleted.

ibm/data_source_ibm_database.go
stevestrutt: Agreed optional. Changed.

ibm/resource_ibm_database.go
stevestrutt: Corrected text. If not set by the user, I believe the Id is locked. There is no default password. I've updated the docs to say that another user must be added.

ibm/resource_ibm_database.go
stevestrutt: Removed extra d.Get("adminpassword")

ibm/resource_ibm_database.go
stevestrutt: Somehow I lost the name update. I've added that back in. Plan cannot be changed as today there is only a single plan "standard". I left plan in the definition to keep it consistent with all other services, but its not really needed.

stevestrutt:
All changes made. A few errors on my part. All tests now run OK in parallel.

ibm/data_source_ibm_database.go
```diff
@@ -0,0 +1,429 @@
...
+
+			"location": {
+				Description: "The location or the environment in which Database instance exists",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+
+			"service": {
+				Description: "The name of the Cloud Internet database service",
+				Type:        schema.TypeString,
```
hkantare: This should be optinal

ibm/data_source_ibm_database.go
```diff
@@ -0,0 +1,429 @@
...
+				Description: "Memory allocation required for cluster",
+				Type:        schema.TypeInt,
+				Computed:    true,
+			},
+			"members_disk_allocation_mb": {
+				Description: "Disk allocation required for cluster",
+				Type:        schema.TypeInt,
+				Computed:    true,
+			},
+			"users": {
```
hkantare: Is this field optional or computed?

ibm/data_source_ibm_database.go
stevestrutt: 'Service' is only optional. Name is the primary key for the Resource Controller and other values can then be used to narrow it down.

ibm/data_source_ibm_database.go
stevestrutt: Users and memory etc are all computed for this data_source. However it lead me to thinking about what was optional and what was not. The schema is OK, but there is a dependency on the provider region parameter being set correctly. The default is us-south and if the instance is not in us-south both the data_source create and import will fail. I've updated the doc to highlight this for both the data_source and resource. Also the error messages if the db instance is not found. The instance is found via the Resource Controller in any region, but the resource specific queries using the ICD API will fail if the provider is not set correctly.

PR #534, opened 2019-03-18
ICD Tagging

stevestrutt:
Implement Global Tagging on ICD resources. Add, Read, Delete and Update tags. Implements framework for adding tagging support to Resource Manager service instances.  
- FYI in vendoring the latest bluemix-go package, govendor appears to have added new records to vendor.json for each bluemix-go sub-package.


ibm/resource_ibm_database.go
```diff
@@ -607,7 +608,20 @@ func resourceIBMDatabaseInstanceRead(d *schema.ResourceData, meta interface{}) e
 		return nil
 	}
 
-	d.Set("tags", instance.Tags)
```
hkantare: Can we move get tags also to separate function like update..so we can use in multiple resources in future

examples/ibm-database/provider.tf
```diff
@@ -0,0 +1,18 @@
...
+variable "bluemix_api_key" {
+  description = "Enter your IBM Cloud API Key, you can get your IBM Cloud API key using: https://console.bluemix.net/iam#/apikeys"
+}
+
+provider "ibm" {
+  softlayer_username = "${var.softlayer_username}"
+  softlayer_api_key  = "${var.softlayer_api_key}"
+  bluemix_api_key    = "${var.bluemix_api_key}"
+  region             = "eu-gb"
+}
```
hkantare: Can we add a READMe file also which explains few lines wt example is about

stevestrutt: Added README.md

ibm/resource_ibm_database.go
stevestrutt: Moved to GetTags in structures.go

PR #549, opened 2019-04-03
Add optional `most_recent` parameter to `ibm_compute_image_template`

arielsalvo:
The data source is pretty useless when using public images from IBM (among many others) that have the same name over the many different versions of the image.
Adding an optional `most_recent` parameter allows the data source to return the highest ID number in the set instead of just the first it encounters while retaining the original behavior if unset or set to false.

This solution is the same as implemented by the [analogous data source in the AWS provider](https://www.terraform.io/docs/providers/aws/d/ami.html)

### DIFF
--- a/ibm/data_source_ibm_compute_image_template.go
+++ b/ibm/data_source_ibm_compute_image_template.go
@@ -26,6 +26,12 @@ func dataSourceIBMComputeImageTemplate() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 			},
+
+			"latest": {
+				Description: "Get the latest id of this image template",
+				Type:        schema.TypeBool,
+				Optional:    true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


ibm/data_source_ibm_compute_image_template.go
```diff
@@ -63,6 +70,13 @@ func dataSourceIBMComputeImageTemplateRead(d *schema.ResourceData, meta interfac
 
 	if len(pubImageTemplates) > 0 {
 		imageTemplate := pubImageTemplates[0]
+		if most_recent {
```
hkantare: Here most_recent means we should filter based on createdate that was created recently..May be you can use same logic which we implemented for datasource of virtual instance https://github.com/IBM-Cloud/terraform-provider-ibm/blob/ad76bfde7d06789e55a4e03f5da52be702f7a247/ibm/data_source_ibm_compute_vm_instance.go#L243

arielsalvo: I reported a bug in softlayer-go to fix this: https://github.com/softlayer/softlayer-go/issues/131
The idea is to get the API to return the information already ordered.
How can I test this with the new version?

jsutton:
@hkantare Is there something in this PR that is preventing it being merged?

arielsalvo:
@jsutton I haven't had time to rework the solution to include the bug-fixing in `softlayer-go`
@hkantare objected to the code I'd provided, so it is not ready to merge

arielsalvo:
No longer relevant

Issue #498, opened 2019-01-29
Cailco configuration file is created with relative path if config_dir is relative


ashishth09:
Terraform v0.11.0
IBM provider v0.14.0

Configuration used

```
data "ibm_resource_group" "resource_group" {
  name = "shard-components"
}


data "ibm_container_cluster_config" "shard_cluster" {
  cluster_name_id   = "xxxxxxxx"
  config_dir        = "."
  region            = "us-south"
  resource_group_id = "${data.ibm_resource_group.resource_group.id}"
  admin             = true
  network           = true
}
```
etcdKeyFile, etcdCertFile, etcdCACertFile are created with relative paths in calicoctl.cfg

I think the configuration should always be created with an absolute path to ease running it from any directory.

hkantare:
closing the issue fixed in 
https://github.com/IBM-Cloud/terraform-provider-ibm/releases/tag/v0.17.0 and later release

PR #565, opened 2019-04-26
#498 fix

smjt:


hkantare:
Did u ran the testcase after the fix?

smjt:
Yes config files are generated with absolute paths now.

PR #572, opened 2019-05-21
Adding changes for ipsec-vpn

UtsavRaychaudhuri:


UtsavRaychaudhuri:
@hkantare Please review the above pr

ibm/resource_ibm_ipsec_vpn.go
```diff
@@ -0,0 +1,466 @@
...
+			},
+			"customer_peer_ip": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"internal_subnet_id": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+			"remote_subnet_id": { //customer subnet id . need atleast one customer subnet id for applying the configuratons
```
hkantare: As per description user can either provide remote_subnet_id or remote_subnet so add ConflictsWith: []string{"remote_subnet"}, respectively to both attributes..

ibm/resource_ibm_ipsec_vpn.go
```diff
@@ -0,0 +1,466 @@
...
+	vpnID, _ := strconv.Atoi(d.Id())
+
+	vpn, err := services.GetNetworkTunnelModuleContextService(sess).
+		Id(vpnID).Mask(ipsecMask).
+		GetObject()
+	if err != nil {
+		return fmt.Errorf("Error retrieving firewall information: %s", err)
+	}
+	d.Set("name", *vpn.Name)
+	d.Set("internal_peer_ip_address", *vpn.InternalPeerIpAddress)
```
hkantare: In Read we should set back values of all attributes it helps during import scenario. Similar to InternalPerrIPAddress set back all attributes remote_subnet_id, phase_one, Phase_two....so on

ibm/resource_ibm_ipsec_vpn.go
```diff
@@ -0,0 +1,466 @@
...
+	sess := meta.(ClientSession).SoftLayerSession()
+	vpnService := services.GetNetworkTunnelModuleContextService(sess)
+
+	vpnID, _ := strconv.Atoi(d.Id())
+
+	// Get billing item associated with the firewall
+	billingItem, err := vpnService.Id(vpnID).GetBillingItem()
+
+	if err != nil {
+		return fmt.Errorf("Error while looking up billing item associated with the firewall: %s", err)
```
hkantare: Log stmt needs to be corrected from firewall to ipsecvpn

ibm/resource_ibm_ipsec_vpn.go
```diff
@@ -0,0 +1,466 @@
...
+			keylife := value["keylife"].(int)
+			vpn.PhaseTwoKeylife = &keylife
+		}
+	}
+	if d.HasChange("preshared_key") {
+		presharedkey := d.Get("preshared_key").(string)
+		vpn.PresharedKey = &presharedkey
+	}
+	if _, ok := d.GetOk("customer_peer_ip"); ok {
+		if d.HasChange("customer_peer_ip") {
```
hkantare: if and else loop has similar code...Do we need loop here?

ibm/resource_ibm_ipsec_vpn.go
UtsavRaychaudhuri: Some attributes are not set during creation. For eg:- if you mention only the datacenter name while creating an IPSecVPN only datacenter,id,name,phase_one and phase_two values are set. Others are not set.These attributes can only be set if the user specifically sets it while creating. So for now i am checking if these values are set and then setting them accordingly. Please tell me if u have a different approach in mind? I will wait for your reply before pushing the code.

hkantare: yes ..We need to check if values are set then only we need to set the attributes accordingly..

ibm/resource_ibm_ipsec_vpn.go
```diff
@@ -0,0 +1,466 @@
...
+			"internal_peer_ip_address": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"phase_one": {
+				Type:     schema.TypeList,
```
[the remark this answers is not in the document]
UtsavRaychaudhuri: Done

ibm/resource_ibm_ipsec_vpn.go
UtsavRaychaudhuri: Done

ibm/resource_ibm_ipsec_vpn.go
UtsavRaychaudhuri: Done. Please review

ibm/resource_ibm_ipsec_vpn.go
UtsavRaychaudhuri: Done. Please review

ibm/resource_ibm_ipsec_vpn.go
UtsavRaychaudhuri: Done. Please review

ibm/resource_ibm_ipsec_vpn.go
```diff
@@ -0,0 +1,537 @@
...
+		GetObject()
+	if err != nil {
+		return fmt.Errorf("Error retrieving firewall information: %s", err)
+	}
+	d.Set("name", *vpn.Name)
+	d.Set("internal_peer_ip_address", *vpn.InternalPeerIpAddress)
+	if vpn.Datacenter != nil {
+		d.Set("datacenter", *vpn.Datacenter.Name)
+	}
+	phaseoneAttributesMap := make([]map[string]interface{}, 0, 1)
```
hkantare: Changes look fine...small chnage Can you move these attributesMap to respective flatten methods in structures.go (flattenIPSecPhaseOne,flattenIPSecPhaseTwo, flattenIPSecAddressTranslation, flattenIPSecCustomerSubnets)

UtsavRaychaudhuri: Done. Please review

UtsavRaychaudhuri:
@hkantare Please review the above pr.

PR #694, opened 2019-08-20
Do not recreate VMs if Security Groups are added/removed

trynity:
It's possible to add/remove security groups easily via the API
and Web UI for IBM Cloud on VMs. There's no need to recreate a VM
resource when changing them via Terraform


acrewdson:
@hkantare From my perspective, the bug this PR attempts to address is a significant shortcoming of the provider as it stands. Not being able to add a security group to an existing VM without destroying the VM in question is something that routinely poses problems for users and impedes effective usage of Terraform to manage IBM Cloud infra.

Can you provide guidance on how this PR should be updated/expanded to address the defect? Any help will be greatly appreciated.

grommir:
@hkantare 
Hi. Any progress on this PR?

PR #700, opened 2019-08-27
Update variables.tf 

EdwardLiaoCJ:
change the cluster default from "cluster" to "mycluster"


EdwardLiaoCJ:
Just a little change

PR #709, opened 2019-08-29
Fix security grp and rule for NextGen support

VaishnaviGopal:


hkantare:
Update security rule doc also with changes in direction "inbound" "outbound"

Issue #715, opened 2019-08-30
HTTPS should be supported for backend_protocol against ibm_lbaas


ganhuang:
### Terraform Version
Terraform v0.11.11

### Affected Resource(s)
Please list the resources as a list, for example:
- ibm_lbaas
- ibm_lbaas_health_monitor

### Terraform Configuration Files
```
/**
 * Generate a Load Balancer (Cloud Load Balancer) to split traffice between the instances.
 */
resource "ibm_lbaas" "load_balancer" {
  name        = "${var.environment}-${var.region}-${var.service}-lb"
  description = "Load balances ${length(split(",", var.instance_addresses))} servers for ${var.service}."
  type        = "${var.lb_visibility}"
  subnets     = ["${data.ibm_network_vlan.private_vlan.subnets.0.id}"]

  protocols = [
    {
      frontend_protocol     = "HTTP"
      frontend_port         = "80"
      backend_protocol      = "HTTP"
      backend_port          = "80"
      load_balancing_method = "${var.load_balancing_method}"
    },
    {
      frontend_protocol     = "HTTPS"
      frontend_port         = "443"
      backend_protocol      = "HTTPS"
      backend_port          = "443"
      load_balancing_method = "${var.load_balancing_method}"
      tls_certificate_id    = "xxxx"
    },
    {
      frontend_protocol     = "HTTPS"
      frontend_port         = "5050"
      backend_protocol      = "HTTPS"
      backend_port          = "5050"
      load_balancing_method = "${var.load_balancing_method}"
      tls_certificate_id    = "xxxx"
    }
  ]
}
```

```
/**
 * Create a health check to ensure the instances are healthy and should receive traffic.
 */
resource "ibm_lbaas_health_monitor" "monitor_443" {
  lbaas_id    = "${ibm_lbaas.load_balancer.id}"
  monitor_id  = "${ibm_lbaas.load_balancer.health_monitors.1.monitor_id}"
  protocol    = "HTTPS"
  port        = "443"
  timeout     = "${var.health_check_timeout}"
  interval    = "${var.health_check_interval}"
  max_retries = "${var.health_check_max_retries}"
  url_path    = "${var.health_check_path}"
}
```

```
/**
 * Create a health check to ensure the instances are healthy and should receive traffic.
 */
resource "ibm_lbaas_health_monitor" "monitor_5050" {
  lbaas_id    = "${ibm_lbaas.load_balancer.id}"
[...]

PR #717, opened 2019-09-02
add HTTPS backend support for lbaas and health monitor

ganhuang:
Proposed fix: https://github.com/IBM-Cloud/terraform-provider-ibm/issues/715


ganhuang:
@hkantare Thanks for the review, I've updated the test case config accordingly.

Is there a place where I can check the failure logs? I don't think it could passed prior the PR.

PR #726, opened 2019-09-05
Fix to stop before deleting the resource "is_instance"

VaishnaviGopal:


ibm/resource_ibm_is_instance.go
```diff
@@ -787,6 +802,34 @@ func isWaitForInstanceDelete(d *schema.ResourceData, meta interface{}) (interfac
...
 
+func isWaitForInstanceActionStop(d *schema.ResourceData, meta interface{}) (interface{}, error) {
+	sess, err := meta.(ClientSession).ISSession()
+	if err != nil {
+		return false, err
+	}
+	instanceC := compute.NewInstanceClient(sess)
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{isInstanceActionStatusRunning, isInstanceActionStatusPending},
```
hkantare: Pending chk for stopping, running, pending

ibm/resource_ibm_is_instance.go
```diff
@@ -787,6 +802,34 @@ func isWaitForInstanceDelete(d *schema.ResourceData, meta interface{}) (interfac
...
+		Target:  []string{isInstanceActionDone},
```
hkantare: target : stopped,failed

ibm/resource_ibm_is_instance.go
```diff
@@ -787,6 +802,34 @@ func isWaitForInstanceDelete(d *schema.ResourceData, meta interface{}) (interfac
...
+		Refresh: func() (interface{}, string, error) {
+			instance, err := instanceC.Get(d.Id())
+			if err != nil {
+				return nil, "", err
+			}
+			if instance.Status == isInstanceActionStatusCompleted {
+				return instance, isInstanceActionDone, nil
```
hkantare: return instance.Status instead of isInstanceActionDone

ibm/resource_ibm_is_instance.go
```diff
@@ -787,6 +802,34 @@ func isWaitForInstanceDelete(d *schema.ResourceData, meta interface{}) (interfac
...
+			}
+			return instance, isInstanceActionStatusRunning, nil
```
hkantare: same as above return instance.Status instead of isInstanceActionStatusRunning

PR #739, opened 2019-09-17
Add resourcefields(name,status,grpname,crn,url) as computed

hkantare:


sakshiag:
@hkantare : Do we need update the readme also for the new attributes?

hkantare:
@sakshiag No this fields are not required to be updated in docs after discussions with Albee...

PR #754, opened 2019-09-23
openshift-job example added

smjtd:


examples/ibm-openshift-job/README.MD
```diff
@@ -0,0 +1,25 @@
+# terraform_cluster_job
```
hkantare: Can you provide  a brief heading what exactly this configuration..Running a job in kubernetes cluster ....

PR #784, opened 2019-10-11
Global Load Balancer terraform example

umarali-nagoor:
ibm_cis_global_load_balancer

Provides a IBM CIS Global Load Balancer resource. This sits in front of a number of defined pools of origins, directs traffic to available origins and provides various options for geographically-aware load balancing. This resource is associated with an IBM Cloud Internet Services instance, a CIS Domain resource and CIS Origin pool resources.


hkantare:
closing this PR may not be required any more

PR #795, opened 2019-10-17
Add support for ssl ciphers and use_system_public_ip_pool for LBass

hkantare:


ibm/data_source_ibm_lbaas.go
```diff
@@ -216,5 +226,11 @@ func dataSourceIBMLbaasRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("protocols", flattenProtocols(result.Listeners))
 	d.Set("health_monitors", flattenHealthMonitors(result.Listeners))
 	d.Set("server_instances", members)
+	d.Set("ssl_ciphers", flattenSSLCiphers(result.SslCiphers))
+	if *result.UseSystemPublicIpPool == 1 {
+		d.Set("use_system_public_ip_pool", true)
+	} else {
+		d.Set("use_system_public_ip_pool", false)
+	}
```
Praveengostu: Can we update the UT for these 2 fields as well

hkantare: Done added testcases for both fields

PR #858, opened 2019-11-07
vpc workerpool and alb test case updated with examples for terraform v0.12

smjtd:


hkantare:
closing the PR already IKS is supported for terraform v0.12

PR #910, opened 2019-11-26
add create/update and delete functionality for image resource

pkrishn1-pk:
This PR is to add support for image resource.

### DIFF
--- a/ibm/resource_ibm_is_image.go
+++ b/ibm/resource_ibm_is_image.go
@@ -1,0 +1,276 @@
+package ibm
+
+import (
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.ibm.com/Bluemix/riaas-go-client/clients/compute"
+	iserrors "github.ibm.com/Bluemix/riaas-go-client/errors"
+)
+
+const (
+	isImageHref            = "href"
+	isImageName            = "name"
+	isImageOperatingSystem = "operating_system"
+	isImageStatus          = "status"
+	isImageVisibility      = "visibility"
+
+	isImageProvisioning     = "provisioning"
+	isImageProvisioningDone = "done"
+	isImageDeleting         = "deleting"
+	isImageDeleted          = "done"
+)
+
+func resourceIBMISImage() *schema.Resource {
+	return &schema.Resource{
+		Create:   resourceIBMISImageCreate,
+		Read:     resourceIBMISImageRead,
+		Update:   resourceIBMISImageUpdate,
+		Delete:   resourceIBMISImageDelete,
+		Exists:   resourceIBMISImageExists,
+		Importer: &schema.ResourceImporter{},
+
+		Schema: map[string]*schema.Schema{
+			isImageHref: {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			isImageName: {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
+			},
+
+			isImageOperatingSystem: {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			isImageStatus: {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			isImageVisibility: {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			ResourceControllerURL: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The URL of the IBM Cloud dashboard that can be used to explore and view details about this instance",
+			},
+
+			ResourceName: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The name of the resource",
+			},
+
+			ResourceCRN: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The crn of the resource",
+			},
+
+			ResourceGroupName: {
+				Type:        schema.TypeString,
+				Computed:    true,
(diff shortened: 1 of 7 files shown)


ibm/resource_ibm_is_image.go
```diff
@@ -0,0 +1,276 @@
...
+
+func resourceIBMISImage() *schema.Resource {
+	return &schema.Resource{
+		Create:   resourceIBMISImageCreate,
+		Read:     resourceIBMISImageRead,
+		Update:   resourceIBMISImageUpdate,
+		Delete:   resourceIBMISImageDelete,
+		Exists:   resourceIBMISImageExists,
+		Importer: &schema.ResourceImporter{},
+
```
hkantare: Add timeout block
```
Timeouts: &schema.ResourceTimeout{
			Create: schema.DefaultTimeout(10 * time.Minute),
			Delete: schema.DefaultTimeout(10 * time.Minute),
		},
```

with default value

ibm/resource_ibm_is_image.go
```diff
@@ -0,0 +1,276 @@
...
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.ibm.com/Bluemix/riaas-go-client/clients/compute"
+	iserrors "github.ibm.com/Bluemix/riaas-go-client/errors"
+)
+
+const (
+	isImageHref            = "href"
```
hkantare: Can this be changed to isImageFile = "file"  instead of href to be inline with cli and api

ibm/resource_ibm_is_image.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	imageC := compute.NewImageClient(sess)
+
+	image, err := imageC.Get(d.Id())
+	if err != nil {
+		return err
+	}
+
+	d.Set(isImageName, image.Name)
+	d.Set(isImageOperatingSystem, image.OperatingSystem)
```
hkantare: Set back all the fields defined in the schema visibility, status, href,...soon

ibm/resource_ibm_is_image.go
```diff
@@ -0,0 +1,276 @@
...
+		iserror, ok := err.(iserrors.RiaasError)
+		if ok {
+			log.Printf("[DEBUG] %s", iserror.Error())
+			if len(iserror.Payload.Errors) == 1 &&
+				iserror.Payload.Errors[0].Code == "not_found" {
+				log.Printf("[DEBUG] returning deleted")
+				return nil, isImageDeleted, nil
+			}
+		}
+		log.Printf("[DEBUG] returning x")
```
hkantare: Remove this log statemment

ibm/resource_ibm_is_image_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.ibm.com/Bluemix/riaas-go-client/clients/compute"
+	"github.ibm.com/Bluemix/riaas-go-client/riaas/models"
+)
+
+func TestAccIBMISImage_basic(t *testing.T) {
+	var image *models.Image
+
+	href := fmt.Sprintf("cos://us-south/vnfimages/BIGIP-15.0.1-0.0.11.qcow2")
```
hkantare: why do we need href and operating declared we already added then as precheck to set the values via enviornmental variables

website/docs/r/is_images.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+* `operating_system` - (Required, string) Description of underlying OS of an image.
+* `tags` - (Optional, array of strings) Tags associated with the image.  
+
+**NOTE**: `Tags` are managed locally and not stored on the IBM Cloud service endpoint at this moment.
+
+## Attribute Reference
+
+The following attributes are exported:
+
+* `id` - The unique identifier of the image.
```
hkantare: Remove href,name, operating_system from Attribute refernce. Add only compute fields here

We have lot of fields like os, file, createdAt which are not declared in scheam

hkantare:
Add an entry for the doc in website/ibm.erb file

ibm/resource_ibm_is_image.go
pkrishn1-pk: added the timeout block..

ibm/resource_ibm_is_image.go
pkrishn1-pk: ok.. setting the same now..

website/docs/r/is_images.html.markdown
pkrishn1-pk: ok.. removed..

ibm/resource_ibm_is_image.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	imageC := compute.NewImageClient(sess)
+
+	image, err := imageC.Get(d.Id())
+	if err != nil {
+		return err
+	}
+
+	d.Set("id", image.ID.String())
+	d.Set("created_at", image.CreatedAt)
```
hkantare: created_at is not define in the schema.

ibm/resource_ibm_is_image.go
```diff
@@ -0,0 +1,330 @@
...
+	imageC := compute.NewImageClient(sess)
+
+	if d.HasChange(isImageName) {
+		name := d.Get(isImageName).(string)
+		_, err := imageC.Update(d.Id(), name)
+		if err != nil {
+			return err
+		}
+	}
+
```
hkantare: Can you add a logic to d.HasChange(isImageTags) ...because tags can be changed after creation

PR #939, opened 2019-12-10
Updated the instance to fix the network order

surajsub:


.gitignore
```diff
@@ -33,3 +33,4 @@ website/vendor
 !command/test-fixtures/**/*.tfstate
 !command/test-fixtures/**/.terraform/
 
+/terraform-provider-ibm
```
hkantare: why we need change in this file?

PR #1125, opened 2020-02-24
added gateway feature for v1 cluster

smjtd:


tpolekhin:
@smjtd @hkantare since this change was merged in we caught an issue with deprecated filed forces new resource for cluster and all related objects like worker_pools
https://github.com/IBM-Cloud/terraform-provider-ibm/issues/1187

PR #1160, opened 2020-03-05
Consider it successful to delete a non-existing FIP

odaira:
Sometimes users need to destroy a resource without refresh `terraform destroy -refresh=false` for various reasons. Even when the resource does not exist any more, the delete operation should be considered successful.

This PR allows deletion of a non-existing FIP.  It is already allowed for a security group rule:
https://github.com/IBM-Cloud/terraform-provider-ibm/blob/b9fb8aeedd852071061c680ec4cfb18f9fde503b/ibm/resource_ibm_is_security_group_rule.go#L286-L296

### DIFF
--- a/ibm/resource_ibm_is_floating_ip.go
+++ b/ibm/resource_ibm_is_floating_ip.go
@@ -281,6 +281,13 @@ func resourceIBMISFloatingIPDelete(d *schema.ResourceData, meta interface{}) err
 
 	err = floatingipC.Delete(d.Id())
 	if err != nil {
+		iserror, ok := err.(iserrors.RiaasError)
+		if ok {
+			if len(iserror.Payload.Errors) == 1 &&
+				iserror.Payload.Errors[0].Code == "service_error" {
+				return nil
+			}
+		}
 		return err
 	}
 


odaira:
Let me know if you have any concern with allowing deletion of a non-existing resource without refresh.

An example case where users need to run `terraform destroy -refresh=false`: https://github.com/hashicorp/terraform/issues/15386

hkantare:
@odaira Intially I was concerned with the change because according to earlier docs of Terraform if its fails to delete any resource we should return back error..but yesterday I was going through latest Terraform docs 
https://www.terraform.io/docs/extend/writing-custom-providers.html#implementing-destroy
```
The destroy function should always handle the case where the resource might already be destroyed (manually, for example). If the resource is already destroyed, this should not return an error. 
```
For Read also they are saying to follow the same logic 
https://www.terraform.io/docs/extend/writing-custom-providers.html#implementing-read
can we update Read Method also?

odaira:
@hkantare I've never thought about Read, but yes, looks like we need to update Read as well to gracefully handle already-deleted resources.

However, I've used Terraform with IBM plugin for already-deleted resources many times, but unless I specified `-refresh=false`, I've never seen any weird error.  I guess even with the current plugin implementation, Terraform gracefully handles an error from Read and simply considers that the resource no longer exists. (This is just my guess.)

PR #1389, opened 2020-04-29
VPC DS migrated acc to new vpc-go-sdk

VaishnaviGopal:


ibm/data_source_ibm_is_region.go
```diff
@@ -36,18 +37,61 @@ func dataSourceIBMISRegion() *schema.Resource {
...
-	d.SetId(region.Name)
-	d.Set(isRegionEndpoint, region.Endpoint)
-	d.Set(isRegionName, region.Name)
-	d.Set(isRegionStatus, region.Status)
+	getRegionOptions := &vpcv1.GetRegionOptions{
+		Name: &name,
+	}
+	region, _, err := sess.GetRegion(getRegionOptions)
+	if err != nil {
+		return nil
```
hkantare, line 89: if err != nil it shd be return err

ibm/data_source_ibm_is_image_test.go
```diff
@@ -9,15 +9,17 @@ import (
 
 func TestAccIBMISImageDataSource_basic(t *testing.T) {
 	resName := "data.ibm_is_image.test1"
+	image := "ubuntu-18.04-amd64" // for classic infrastructure
```
hkantare, line 12: Declare some environmental variable how we did for other SL_INSTANCE_PROFILE

Issue #1430, opened 2020-05-11
Intermittent errors while creating dns service records


j4zzcat:
2020-05-11T17:54:29.204Z [DEBUG] plugin.terraform-provider-ibm_v1.5.0: 2020/05/11 17:54:29 Configuring SoftLayer Session with API key
2020-05-11T17:54:29.204Z [DEBUG] plugin.terraform-provider-ibm_v1.5.0: 2020/05/11 17:54:29 Configuring IBM Cloud Session with API key
2020-05-11T17:54:31.430Z [DEBUG] plugin.terraform: file-provisioner (internal) 2020/05/11 17:54:31 [ERR] Checkpoint error: Get https://checkpoint-api.hashicorp.com/v1/check/terraform?arch=amd64&os=linux&signature=77a380de-a584-50e8-aed4-8b00d10c8e03&version=0.12.24: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
2020-05-11T17:54:31.516Z [DEBUG] plugin.terraform: remote-exec-provisioner (internal) 2020/05/11 17:54:31 [ERR] Checkpoint error: Get https://checkpoint-api.hashicorp.com/v1/check/terraform?arch=amd64&os=linux&signature=77a380de-a584-50e8-aed4-8b00d10c8e03&version=0.12.24: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
2020-05-11T17:54:31.574Z [DEBUG] plugin.terraform: local-exec-provisioner (internal) 2020/05/11 17:54:31 [ERR] Checkpoint error: Get https://checkpoint-api.hashicorp.com/v1/check/terraform?arch=amd64&os=linux&signature=77a380de-a584-50e8-aed4-8b00d10c8e03&version=0.12.24: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
2020-05-11T17:54:36.560Z [DEBUG] plugin.terraform-provider-ibm_v1.5.0: 2020/05/11 17:54:36 the apiendpoint url for power is us-south.power-iaas.cloud.ibm.com
ibm_dns_resource_record.srv_records[0]: Destroying... [id=431e4ec8-f9bc-4014-85fb-36695e441cc2/blue.cloud:9c966a8a-a821-4049-bbb8-58a74fdb567a/SRV:17005065-28c4-42ec-80ac-6d9f14034738]
[...]

PR #1490, opened 2020-05-26
Changes to lock resource record id and zone id

MalarvizhiK:
Fix for defect #1430


ibm/resource_ibm_private_dns_resource_record.go
```diff
@@ -355,12 +359,16 @@ func resourceIBMPrivateDNSResourceRecordRead(d *schema.ResourceData, meta interf
 }
 
 func resourceIBMPrivateDNSResourceRecordUpdate(d *schema.ResourceData, meta interface{}) error {
+	id_set := strings.Split(d.Id(), "/")
+	mk := "private_dns_resource_record_" + id_set[2]
```
hkantare, line 363: I don't think the lock should be on id_set[2] which is the resource record type ID...Its unique for each resource.

It should be also base don zone_id (id_set[1])

MalarvizhiK: I am testing whether update works with zone id or not.

PR #1517, opened 2020-06-02
added roks on vpc example

Anil-CM:


hkantare:
closing this PR ..The example is added as part of other example

PR #1534, opened 2020-06-09
Data source support for vpc instance, instances

VaishnaviGopal:


ibm/provider_test.go
```diff
@@ -311,6 +311,8 @@ func init() {
 	if isImage == "" {
 		isImage = "fc538f61-7dd6-4408-978c-c6b85b69fe76" // for classic infrastructure
 		// isImage = "r006-ed3f775f-ad7e-4e37-ae62-7199b4988b00" // for next gen infrastructure
+		// isImage := "a7a0626c-f97e-4180-afbe-0331ec62f32a" // classic windows machine: ibm-windows-server-2012-full-standard-amd64-1
```
hkantare, line 314: Create a seperate variable for isWinImage and use that in this testcase...

PR #1617, opened 2020-07-01
Regard all the status other than stopped as stopping

odaira:
`isWaitForInstanceActionStop()` regards only `running`, `pending`, and `stopping` as pending status.  However, RIAS API defines more other status:
```
status*	string
The status of the virtual server instance
Enum:
[ failed, paused, pausing, pending, restarting, resuming, running, starting, stopped, stopping ]
```

Specifically, when trouble occurs during provisioning, an instance can remain in the `starting` status.  When TF tries to destroy such an instance, it will throw an error about the unrecognized status, which would confuse the user:
```
Error: unexpected state 'starting', wanted target 'stopped, failed, '. last error: %!s(<nil>)
```

I think `isWaitForInstanceActionStop()` should regard all the status other than `stopped` (and `failed`) as stopping.

### DIFF
--- a/ibm/resource_ibm_is_instance.go
+++ b/ibm/resource_ibm_is_instance.go
@@ -1997,7 +1997,7 @@ func isWaitForInstanceDelete(instanceC *vpcv1.VpcV1, d *schema.ResourceData, id 
 func isWaitForClassicInstanceActionStop(instanceC *vpcclassicv1.VpcClassicV1, d *schema.ResourceData, meta interface{}, id string) (interface{}, error) {
 
 	stateConf := &resource.StateChangeConf{
-		Pending: []string{isInstanceStatusRunning, isInstanceStatusPending, isInstanceActionStatusStopping},
+		Pending: []string{isInstanceActionStatusStopping},
 		Target:  []string{isInstanceActionStatusStopped, isInstanceStatusFailed, ""},
 		Refresh: func() (interface{}, string, error) {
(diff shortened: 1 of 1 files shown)


hkantare:
@odaira Does the user may not get wrong status of instance saying stopping where as the actual status will be starting, pending so on

odaira:
@hkantare well, yes, with this change, the user would see a timeout error like this, while the actual status is starting, etc.
```
Error: timeout while waiting for state to become 'stopped, failed, ' (last state: 'stopping', timeout: 20m0s)
```
I agree that this is somewhat confusing, too.

However, in a sense, this behavior is consistent with an error during provisioning (`isWaitForInstanceAvailable()`)or deleting (`isWaitForInstanceDelete()`).  The user can receive an error like this, while there is no such status as `deleting` defined by the API.
```
Error: timeout while waiting for state to become 'done, ' (last state: 'deleting', timeout: 30m0s)
```

If the pseudo status of `stopping` seems confusing with the actual status of `stopping`, we can come up with new wording....

odaira:
@hkantare What do you think would be the best way to handle this situation?  Do you think the user should see the actual status in the error message?

hkantare:
@odaira Yes as per my understanding the user should the current status of the resource...so when they compare in ui or cli they will get the same error

hkantare:
closing this PR due to inactivity

PR #1633, opened 2020-07-06
Defect 1624

MalarvizhiK:
Changes for defect 1624 - resource ibm_dns_zone instance_id takes the ibm_resource_instance.x.guid not the id

### DIFF
--- a/ibm/resource_ibm_private_dns_permitted_network.go
+++ b/ibm/resource_ibm_private_dns_permitted_network.go
@@ -2,7 +2,6 @@ package ibm
 
 import (
 	"fmt"
-	"log"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


ibm/resource_ibm_private_dns_permitted_network.go
```diff
@@ -183,7 +179,7 @@ func resourceIBMPrivateDnsPermittedNetworkExists(d *schema.ResourceData, meta in
 	_, response, err := sess.GetPermittedNetwork(getPermittedNetworkOptions)
 	if err != nil {
 		if response != nil && response.StatusCode == 404 {
-			return false, nil
+			return false, fmt.Errorf("Error fetching pdns permitted network:%s\n%s", err, response)
 		}
```
hkantare, line 183: Don't return err if response != nil && response.StatusCode == 404 {
}
keep the the old code only
return false, nil

ibm/resource_ibm_private_dns_resource_record.go
```diff
@@ -517,7 +511,7 @@ func resourceIBMPrivateDNSResourceRecordExists(d *schema.ResourceData, meta inte
 
 	if err != nil {
 		if response != nil && response.StatusCode == 404 {
-			return false, nil
+			return false, fmt.Errorf("Error fetching pdns resource record:%s\n%s", err, response)
```
hkantare, line 514: similarly don;t change here we are retuning nil

PR #1655, opened 2020-07-08
Cloud functions

anilkumarnagaraj:


ibm/resource_ibm_function_action.go
```diff
@@ -145,17 +160,67 @@ func resourceIBMFunctionAction() *schema.Resource {
 				Computed:    true,
 				Description: "All paramters set on action by user and those set by the IBM Cloud Function backend/API.",
 			},
+			"action_id": {
+				Type:     schema.TypeString,
+				Computed: true,
```
hkantare, line 165: You defined the action_id but d.Set the value is not set  simialrly for other resources also can you check

website/docs/r/function_trigger.html.markdown
```diff
@@ -94,6 +97,7 @@ The following arguments are supported:
 The following attributes are exported:
 
 * `id` - The ID of the new trigger.
```
hkantare, line 99: Change all resource docs to say the id is combination of namespace/respectiveresourceID
Add respective new attribute like "action_id"..so on 

Add a import section block in resources blocks which describes how you do import ...Refer some old docs like is_lb_pool docs for more etails

hkantare:
Make the provider level argument `function_namespace` in provider.go as deprecated  and respective message (This field will be deprecated soon .

PR #1686, opened 2020-07-14
add extensions as exported parameter in resource instance

JunliWang:
We(Event Streams) is working on Terraform support, and we would like to define the resource and user scenario in this way
1. we will expose `ibm_event_streams_topic` as a resource for user to create under an instancce, and use `ibm_resource_instance` as it is.
2. we also need to expose a connection string at instance level for customer's application to connect to. The proposal is add `extensions` as an output parameters in `ibm_resource_instance`, then our service broker will store this connection string in that map, then customer application can use `lookup(ibm_resource_instance.es_instance.extensions, "bootstrap.server", "")` to get it. An alternative way(as database implemented as such) is defining a resource `ibm_event_streams` that is equivalent to `ibm_resource_instance` and export the `bootstrap.server` in that resource, but it has to duplicate all the CRUD operations that `ibm_resource_instance` already has done. Thus I do not prefer this way.
While exporting `extensions` from `ibm_resource_instance` gives more flexibility for all services who want to export some properties for their consumers at instance level.

a dependent PR is done: https://github.com/IBM-Cloud/bluemix-go/pull/229
 
```hcl
variable "ibmcloud_api_key" {}
variable "es_reader_api_key" {}

provider "ibm" {
  ibmcloud_api_key = var.ibmcloud_api_key
  generation       = 1
}

data "ibm_resource_group" "group" {
  name = "Default"
}

resource "ibm_resource_instance" "es_instance" {
  name              = "terraform-integration"
  service           = "messagehub"
  plan              = "standard"
  location          = "us-south"
  resource_group_id = data.ibm_resource_group.group.id
}

resource "ibm_event_streams_topic" "es_topic" {
  resource_instance_id = ibm_resource_instance.es_instance.id
  name                 = "my-es-topic"
  partitions           = 1
  config = {
    "cleanup.policy"  = "compact,delete"
    "retention.ms"    = "86400000"
    "retention.bytes" = "1073741824"
    "segment.bytes"   = "536870912"
  }
}

// assume customer wants to deploy a kafka consumer application connecting to an Event Streams instance
resource "kafka_consumer_app" "my_kafka_app" {
  bootstrap_server = lookup(ibm_resource_instance.es_instance.extensions, "bootstrap.server", "")
  topic            = ibm_event_streams_topic.es_topic.name
  apikey           = var.es_reader_api_key // not ibmcloud_api_key(platform apikey), but a service-id-api-key with reduced privileges,eg.Reader role.
}
```


hkantare:
Hi
Can you update the docs for resource also website/docs/t

Issue #1681, opened 2020-07-14
VPC IKS workers upgrade should be optional


tpolekhin:
When upgrading IKS version via terraform it also replaces all worker nodes in the cluster to upgrade them to the specified master version as well. This is a destructive behavior that brings your entire cluster down for an upgrade period.

Upgrading workers together with the master upgrade should be optional
https://github.com/IBM-Cloud/terraform-provider-ibm/blob/master/ibm/resource_ibm_container_vpc_cluster.go#L475

mschenk42:
I agree, the automatic upgrade for the workers needs to be optional.  There is no way we can run this on a production cluster, especially with stateful services that need some care when upgrading. Can we help add this functionality?  We have experienced Go developers that would like to help.

hkantare:
@mschenk42 yes you can gohead and contribute may be we need to introduce a bool argument

mschenk42:
Thanks, will do.  Currently testing change.  Will be submitting a PR soon.

hkantare:
Fixed in latest release
https://github.com/IBM-Cloud/terraform-provider-ibm/releases/tag/v1.9.0

tpolekhin:
Fixed by #1694

PR #1694, opened 2020-07-16
flag to control if workers are reloaded

mschenk42:
This PR is in reference to  #1681 

I ported over the feature from the `ibm_container_cluster resource`.


mschenk42:
Hi @hkantare, let me know if you need anything more for this PR. I did test locally by creating an IKS cluster and then upgrading it using the updated provider.  The `ibm_container_cluster` resource didn't seem to have specific tests for this flag.  However if you need us to add some tests we can.  We just need guidance on what you would be looking for. Thanks.

hkantare:
Thanks for PR...Can you please update doc file under (website/docs/r/container_vpc_cluster.html.markdown) folder

PR #1698, opened 2020-07-16
Handling Interface nil pointer exception for remote argument

umarali-nagoor:
In VPC, argument "remote" could be a ip address, CIDR block or security group ID.  It comes as map[string]interface{}. 

Added a check over the remote before looping through the map.  Based on the key ( could be id, address or cidr_block ) populating the remote value.

Screenshot 2020-07-16 at 5 17 52 PM
Screenshot 2020-07-16 at 5 17 59 PM
Screenshot 2020-07-16 at 5 27 26 PM

### DIFF
--- a/ibm/resource_ibm_is_vpc.go
+++ b/ibm/resource_ibm_is_vpc.go
@@ -654,6 +654,7 @@ func classicVpcGet(d *schema.ResourceData, meta interface{}, id string) error {
 	securityGroupList := make([]map[string]interface{}, 0)
 
 	for _, group := range sgs.SecurityGroups {
+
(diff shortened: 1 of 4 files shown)


ibm/resource_ibm_is_vpc.go
```diff
@@ -200,6 +200,7 @@ func resourceIBMISVPC() *schema.Resource {
 			isVPCSecurityGroupList: {
 				Type:     schema.TypeList,
 				Computed: true,
+				Optional: true,
```
hkantare, line 203: why optional field attribute its just computed

PR #1707, opened 2020-07-19
set labels on default worker pool

mschenk42:
This allows for setting the default worker labels when using the `ibm_container_vpc_cluster` resource to create a cluster.


mschenk42:
Hi @hkantare - is there anything more I need to do when submitting a PR?  Just wondering how this eventually gets incorporated into the main branch. Thanks.

hkantare:
Can we set the labels in Read method

mschenk42:
@hkantare I see, miss that one,  good catch, I added the reading of the worker pool labels.

Issue #1262, opened 2020-04-02
ibm_is_network_acl Validate name input field of ACL Rules


seafre:
### Terraform Version
```
terraform -v
Terraform v0.12.24
+ provider.ibm v1.2.6
+ provider.null v2.1.2
+ provider.tls v2.1.1
```

### Affected Resource(s)
Please list the resources as a list, for example:
- ibm_is_network_acl

### Terraform Configuration Files
```hcl
resource "ibm_is_network_acl" "vpc_subnet_acl" {
  name = "vpc-subnet-acl"
  vpc  = ibm_is_vpc.vpc.id

  rules {
    name        = "icmp_outbound"
    action      = "allow"
    source      = "0.0.0.0/0"
    destination = "0.0.0.0/0"
    direction   = "outbound"
    icmp {
      code = 1
      type = 1
    }
```

### Expected Behavior
`Error: "name" ("icmp_outbound") of "rules" on resource "ibm_is_network_acl" should contain only lowercase alphanumeric,dash and should begin with lowercase character`

### Actual Behavior
```
ibm_is_network_acl.vpc_subnet_acl: Destroying... [id=xxxx-xxxx]
ibm_is_network_acl.vpc_subnet_acl: Destruction complete after 1s
ibm_is_network_acl.vpc_subnet_acl: Creating...

Error: {"errors":[{"code":"bad_field","message":"Invalid json payload provided","target":{"name":"name","type":"field"}}],"trace":"xxxx-xxxx"}

  on vsi.tf line 127, in resource "ibm_is_network_acl" "vpc_subnet_acl":
 127: resource "ibm_is_network_acl" "vpc_subnet_acl" {
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
Fixed by replacing underscore `_` with dash `-`, as performed in all other names, but error message does not suggest where to perform fix to Terraform Configuration File

### References
Similar to other issue regarding validation of name input fields:
- GH-1090
- Commit 8d6203b168c910b5b27e8a880981a66ab3389b36

PR #1799, opened 2020-08-13
 added validate func to vpc nwacl,Fix issue #1262

VaishnaviGopal:


ibm/resource_ibm_is_networkacls_test.go
```diff
@@ -167,7 +167,7 @@ func testAccCheckIBMISNetworkACLConfig() string {
 func testAccCheckIBMISNetworkACLConfig1() string {
 	return fmt.Sprintf(`
 	resource "ibm_is_vpc" "testacc_vpc" {
-		name = "vpctest"
+		name = "tf-nwacl-vpc"
```
hkantare, line 170: why is the name changed?

PR #1867, opened 2020-09-04
Added resource for ibm_pi_snapshot + cleaned up pi_instance

surajsub:


yussufsh:
@hkantare Can we expect this in the next release? Thanks.

ibm/resource_ibm_pi_network_port.go
```diff
@@ -44,14 +45,9 @@ func resourceIBMPINetworkPort() *schema.Resource {
...
-			/*
-			 "description": "",
-			    "ipAddress": "172.31.96.93",
-			    "macAddress": "fa:16:3e:30:b9:64",
-			    "portID": "6c9d0e42-73f3-492f-840d-3d4a7a573014",
-			*/
+			//Computed Attributes
 
-			helpers.PINetworkPortIPAddress: {
+			"ipaddress": {
```
hkantare, line 50: The required and option fields for all power resources are having this naming convnetion helpers.PINetworkPortIPAddress (pi_XXXX) I think please follow same here also

ibm/data_source_ibm_pi_image.go
```diff
@@ -47,6 +47,10 @@ func dataSourceIBMPIImage() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"storage_type": {
```
hkantare, line 50: Add this storage_type  attribute to the docs of both resource and datasources pi_image

ibm/resource_ibm_pi_network_port_attach.go
```diff
@@ -40,18 +40,21 @@ func resourceIBMPINetworkPortAttach() *schema.Resource {
...
 			helpers.PINetworkName: {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Network Name - This is the subnet name  in the Cloud instance",
 			},
 
 			"description": {
-				Type:     schema.TypeString,
```
hkantare, line 53: Description is optional field...can we add the helper string pi_description and in create don't set the default description if it's not set by user
```
if d.Get(description) == "" {
		description = "Testing from Terraform"
	}
```

PR #1895, opened 2020-09-21
updated for lb poolist

pinky-bhargava:


examples/ibm-is-vpc/variables.tf
```diff
@@ -18,3 +18,5 @@ variable "profile" {
   default = "bc1-2x8"
 }
 
+variable "lb_name" {
```
hkantare, line 21: Don't take name of the load balancer from user..Use the load balancer created in the example it self

examples/ibm-is-vpc/main.tf
```diff
@@ -318,4 +318,6 @@ resource "ibm_is_public_gateway" "publicgateway1" {
 data "ibm_is_vpc" "vpc1" {
   name = ibm_is_vpc.vpc1.name
 }
-
+data "ibm_is_lb" "test_lb" {
+name = var.lb_name
```
hkantare, line 322: name = ibm_is_lb.name

PR #1906, opened 2020-09-23
vpc-autoscale fixes

Anil-CM:
This PR has the fix for 
1. autoscale resource validator functions.
2. Handling the instance count update scenario for managed instance group.

### DIFF
--- a/ibm/resource_ibm_is_instance_group.go
+++ b/ibm/resource_ibm_is_instance_group.go
@@ -52,7 +52,8 @@ func resourceIBMISInstanceGroup() *schema.Resource {
 
 			"instance_count": {
 				Type:         schema.TypeInt,
-				Required:     true,
+				Optional:     true,
+				Default:      0,
 				ValidateFunc: InvokeValidator("ibm_is_instance_group", "instance_count"),
 				Description:  "The number of instances in the instance group",
 			},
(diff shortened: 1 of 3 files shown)


ibm/resource_ibm_is_instance_group.go
```diff
@@ -358,6 +371,41 @@ func resourceIBMISInstanceGroupDelete(d *schema.ResourceData, meta interface{})
 	return nil
 }
 
+func isInstanceGroupManagerEnabled(d *schema.ResourceData, meta interface{}) (bool, error) {
```
hkantare, line 374: Are you using this function any where?

website/docs/r/is_instance_group.html.markdown
```diff
@@ -69,7 +69,7 @@ The following arguments are supported:
 
 * `name` - (Required, Forces new resource, string) The name of the instance group.
 * `instance_template` - (Required, Forces new resource, string) The ID of the instance template to create the instance group.
-* `instance_count` - (Required, int) The number of instances to be created under the instance group.
+* `instance_count` - (Optional, int) The number of instances to be created under the instance group. Default is set to 0.
```
hkantare, line 72: Add a comment here saying the instance_count can be updated only when the manager is disabled

PR #1916, opened 2020-09-24
added datasource for lb list

pinky-bhargava:


ibm/data_source_ibm_is_lbs.go
```diff
@@ -0,0 +1,507 @@
...
+		Read: dataSourceIBMISLBSRead,
+		Schema: map[string]*schema.Schema{
+			loadBalancers: {
+				Type:        schema.TypeList,
+				Description: "Collection of load balancers",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						ID: {
+							Type:     schema.TypeString,
```
hkantare, line 32: Small letter "id"

ibm/data_source_ibm_is_lbs_test.go
```diff
@@ -0,0 +1,53 @@
...
+		name            = "%s"
+		vpc             = ibm_is_vpc.testacc_vpc.id
+		zone            = "%s"
+		ipv4_cidr_block = "%s"
+	  }
+	  resource "ibm_is_lb" "testacc_lb" {
+		name    = "%s"
+		subnets = [ibm_is_subnet.testacc_subnet.id]
+	  }
+      data "ibm_is_lbs" "test_lbs" {
```
hkantare, line 51: Split the testcase into two section first create the lb and then run datasource because the datasource executes before resources

website/docs/d/is_lbs.html.markdown
```diff
@@ -0,0 +1,60 @@
...
+```hcl
+data "ibm_is_lbs" "ds_lbs" {
+ }
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` -  Name of the loadbalancer.
```
hkantare, line 25: We don't have name argument ..Remove it

PR #1960, opened 2020-10-13
Doc Update: Cluster Addons Docs

kavya498:


geemurth:
Thanks for the update, kavya. I have added in the document.

PR #1967, opened 2020-10-15
Added expiration date field to ibm_kms_key resource

Sidd-kp:
Description : Added expiration date functionality to key creation in ibm_kms_key and made appropriate changes in test file for testing.

### DIFF
--- a/ibm/resource_ibm_kms_key.go
+++ b/ibm/resource_ibm_kms_key.go
@@ -99,7 +99,7 @@ func resourceIBMKmskey() *schema.Resource {
 				Optional:    true,
 				Description: "The date the key material expires. The date format follows RFC 3339. You can set an expiration date on any key on its creation. A key moves into the Deactivated state within one hour past its expiration date, if one is assigned. If you create a key without specifying an expiration date, the key does not expire",
 				ForceNew:    true,
-				Default:     "Default",
+				Default:     "nil",
 			},
 			ResourceName: {
 				Type:        schema.TypeString,
@@ -195,7 +195,8 @@ func resourceIBMKmsKeyCreate(d *schema.ResourceData, meta interface{}) error {
 	expiration_string := d.Get("expiration_date").(string)
 	var expiration_time time.Time
 	expiration := &expiration_time
-	if expiration_string != "Default" {
+	if expiration_string != "nil" {
+		// parse string to required time format
 		expiration_time, err = time.Parse(time.RFC3339, expiration_string)
 		if err != nil {
 			return fmt.Errorf("Invalid time format: %s", err)
--- a/ibm/resource_ibm_kms_key.go
+++ b/ibm/resource_ibm_kms_key.go
@@ -99,7 +99,6 @@ func resourceIBMKmskey() *schema.Resource {
 				Optional:    true,
 				Description: "The date the key material expires. The date format follows RFC 3339. You can set an expiration date on any key on its creation. A key moves into the Deactivated state within one hour past its expiration date, if one is assigned. If you create a key without specifying an expiration date, the key does not expire",
 				ForceNew:    true,
-				Default:     "nil",
 			},
 			ResourceName: {
 				Type:        schema.TypeString,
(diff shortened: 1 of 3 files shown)


dinesh-venkatraman (changes requested):
Since expiration_date is a date field, the Default value has to be a default date or as a null. Setting a string value is not necessary. You can check if any default value is set in KMK or set this a null

Sidd-kp:
So the default value is null but we cannot assign a string as null. The word Default is just a placeholder. Later in the code I have handled it and am sending a null value if 'Default' is not set to any date.

ibm/resource_ibm_kms_key.go
```diff
@@ -94,6 +94,13 @@ func resourceIBMKmskey() *schema.Resource {
 				Computed:    true,
 				Description: "Crn of the key",
 			},
+			"expiration_date": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The date the key material expires. The date format follows RFC 3339. You can set an expiration date on any key on its creation. A key moves into the Deactivated state within one hour past its expiration date, if one is assigned. If you create a key without specifying an expiration date, the key does not expire",
+				ForceNew:    true,
+				Default:     "Default",
```
hkantare, line 102: Don't set any Default value....since the API doesn't  have any Default value

ibm/resource_ibm_kms_key.go
```diff
@@ -185,13 +192,23 @@ func resourceIBMKmsKeyCreate(d *schema.ResourceData, meta interface{}) error {
 	kpAPI.Config.InstanceID = instanceID
 	name := d.Get("key_name").(string)
 	standardKey := d.Get("standard_key").(bool)
-
+	expiration_string := d.Get("expiration_date").(string)
```
hkantare, line 195: For optional fields use d.GetOk()

ibm/resource_ibm_kms_key.go
```diff
@@ -185,13 +192,23 @@ func resourceIBMKmsKeyCreate(d *schema.ResourceData, meta interface{}) error {
...
+	var expiration_time time.Time
+	expiration := &expiration_time
+	if expiration_string != "Default" {
```
hkantare, line 198: So all these logic can be moved  loop d.GetOk("XXXXX");ok {
}

ibm/resource_ibm_kms_key_test.go
```diff
@@ -125,7 +137,7 @@ func testAccCheckIBMKmsResourceRootkeyWithCOSConfig(instanceName, KeyName, cosIn
 		key_protect          = ibm_kms_key.test.id
 	}
 	
-`, instanceName, KeyName, cosInstanceName, bucketName)
+`, instanceName, KeyName, expirationdate, cosInstanceName, bucketName)
 }
```
hkantare, line 141: Don't change all existing testcase because we are missing the flow of expiration  nil

hkantare (commented):
Add the argument in doc 
website/docs/r/kms_key.html.markdown

ibm/resource_ibm_kms_key.go
Sidd-kp: Ok. Will do so.

ibm/resource_ibm_kms_key.go
Sidd-kp: Ok

ibm/resource_ibm_kms_key.go
Sidd-kp: Sure that would be cleaner.

ibm/resource_ibm_kms_key_test.go
Sidd-kp: Which ones should I revert?

Sidd-kp: Or would I need to write my own test cases?

hkantare: Let the old testcase remain same..write a new testcase

Sidd-kp: Ok.

Sidd-kp:
Screenshot 2020-10-20 at 3 26 42 PM
Screenshot 2020-10-20 at 3 27 08 PM

PR #1983, opened 2020-10-21
Fix: CIS - Domain Settings mobile redirect crash

KumarGanesanIBM:


ibm/resource_ibm_cis_tls_settings_test.go
```diff
@@ -25,7 +25,6 @@ func TestAccIBMCisTLSSettings_Basic(t *testing.T) {
 			{
 				Config: testAccCheckCisTLSSettingsConfigBasic2("test", cisDomainStatic),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(name, "tls_1_3", "zrt"),
```
kavya498, line 28: WHy was it removed instead of making it on?

KumarGanesanIBM: If we set tls_1_3 version to be "on", we are not sure that, it will set as "on" or "zrt".  that's why I removed explicit checking.

PR #1988, opened 2020-10-25
fix: IAM access groups test CheckDestroy method

mkistler:
This PR fixes a minor error in the "CheckDestroy" method of the IAM Access Groups resource test file.

According to the [Terraform docs](https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html#checkdestroy), the CheckDestroy method should:
> return an error if any resources remain.

As originally coded, the CheckDestroy method would not return an error if `err == nil`, but a get that returns `err == nil` indicates that the requested resource still exists (and was returned).

### DIFF
--- a/ibm/resource_ibm_iam_access_group_test.go
+++ b/ibm/resource_ibm_iam_access_group_test.go
@@ -94,7 +94,9 @@ func testAccCheckIBMIAMAccessGroupDestroy(s *terraform.State) error {
 		// Try to find the key
 		_, _, err := accClient.AccessGroup().Get(agID)
 
-		if err != nil && !strings.Contains(err.Error(), "404") {
+		if err == nil {
+			return fmt.Errorf("Access group still exists: %s", rs.Primary.ID)
+		} else if !strings.Contains(err.Error(), "404") {
 			return fmt.Errorf("Error waiting for access group (%s) to be destroyed: %s", rs.Primary.ID, err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


mkistler:
It appears that this error is actually pervasive throughout the test code, and even in the example Destroy method shown in the  `CONTRIBUTING.md`.  I'll amend this PR to address all the instances of this error.

ibm/resource_ibm_iam_access_group_test.go
```diff
@@ -94,7 +94,7 @@ func testAccCheckIBMIAMAccessGroupDestroy(s *terraform.State) error {
 		// Try to find the key
 		_, _, err := accClient.AccessGroup().Get(agID)
 
-		if err != nil && !strings.Contains(err.Error(), "404") {
+		if err == nil || !strings.Contains(err.Error(), "404") {
 			return fmt.Errorf("Error waiting for access group (%s) to be destroyed: %s", rs.Primary.ID, err)
```
mkistler, line 98: We need to guard against referencing `err` here, since it might be `nil`.  I'll fix that.

mkistler:
I have updated the IAM Access Groups resource test and the `CONTRIBUTING.md` to use what I think is the correct code pattern.  Please review and give feedback.  If there is agreement on this pattern I will make the necessary changes in the other files either another commit in the PR or as a separate PR.

PR #1962, opened 2020-10-13
Updated the network_port to include the public_ip

surajsub:

PR #1999, opened 2020-10-29
Updated the network_port to include the public_ip

### LINKED ISSUE #1930: Provide a way to reserve external IP without creating an instance
Need to provide a way to reserve network addresses on a public network using `ibm_pi_network_port`. Currently, when we create a network_port we can get only a private network address. Our use case needs to know the external/public address associated with that public network port.

### Affected Resource(s)
- ibm_pi_network
- ibm_pi_network_port

### Expected Behavior
When reserving a network port on a public network we should be able to read external IP address as well.

### Actual Behavior
The network port only returns the IP address and no external IP is returned.

### Use case
Our use case is to reserve a network port on a public network. The external IP of the network port will be used as a failover IP in accessing HA services running on multiple VMs with public network.

yussufsh:
Fixes #1930

Signed-off-by: Yussuf Shaikh <yussuf.shaikh@ibm.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.12
 require (
 	github.com/Bowery/prompt v0.0.0-20190916142128-fa8279994f75 // indirect
 	github.com/IBM-Cloud/bluemix-go v0.0.0-20201019071904-51caa09553fb
-	github.com/IBM-Cloud/power-go-client v1.0.48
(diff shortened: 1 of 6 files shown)


yussufsh:
@surajsub @hkantare 
As requested have create PR with port changes only from #1962.

PR #2005, opened 2020-11-04
Changes for enhanced security in DL Gateway

MalarvizhiK:
Changes for enhanced security in Direct Link Gateway


MalarvizhiK:
I corrected my code after review comments from @hkantare.

PR #2047, opened 2020-11-19
Handle the delete instance flow correctly

### LINKED ISSUE #2046: ibm_pi_instance delete flow does not wait till completion
When the ibm_pi_instance is deleted the control does not wait till the process of deletion is completed. This leads to errors where we :
1. Cannot run apply after tainting the ibm_pi_instance resource with error: 
2. Public network destroy fails with error: ```One or more ports have an IP allocation from this subnet```

### Affected Resource(s)
Please list the resources as a list, for example:
- ibm_pi_instance 
- ibm_pi_network

### Debug Output
Error during destroy when instance is deleted but public network fails. A re-destroy after some seconds completes successfully:
```
Error: {"description":"an error has occurred; please try again: unable to delete subnet for network 1ec09a75-bdd9-44f............7992f for cloud instance 01611c7............daec: Expected HTTP response code [] when accessing [DELETE https://192.168.16.136:9696/v2.0/subnets/bde8599a-............1-e271a020d2eb], but got 409 instead\n{\"NeutronError\": {\"message\": \"Unable to complete operation on subnet bde8599a-b72b-49e3-8771-e271a020d2eb: One or more ports have an IP allocation from this subnet.\", \"type\": \"SubnetInUse\", \"detail\": \"\"}}","error":"internal server error"}
```

Error during taint and then apply:
```
Error: failed to provision {"description":"bad request: invalid name server name already exists for cloud-instance","error":"bad request"}
```

### Expected Behavior

- The destroy command should work fine when using instance and public network resources.
- The taint on ibm_pi_instance should work properly by deleting and re-recreating the instance.

### Actual Behavior

- The destroy command fails at random times with error specified above,
- The taint on ibm_pi_instance fails always.

yussufsh:
Before the instance destroy was returning immediately leading to failures in deleting the dependent resources such as network used by the instance. This will also fix the resource taint issue with smooth apply after taint.

Fixes #2046

Signed-off-by: Yussuf Shaikh <yussuf@us.ibm.com>

### DIFF
--- a/ibm/resource_ibm_pi_instance.go
+++ b/ibm/resource_ibm_pi_instance.go
@@ -740,7 +740,7 @@ func isPIInstanceDeleteRefreshFunc(client *st.IBMPIInstanceClient, id, powerinst
 			return pvm, helpers.PIInstanceNotFound, nil
 
 		}
-		return pvm, helpers.PIInstanceNotFound, nil
+		return pvm, helpers.PIInstanceDeleting, nil
 
 	}
 }


yussufsh:
@hkantare @surajsub

PR #2051, opened 2020-11-19
Added input to ibm container data source in order to skip listing bounded services

jfeddern:
Hi there, 
a lot of services are bound to our Kubernetes cluster. Everytime we run the Terraform scripts the ibm_container_cluster data source is executed which fetches all bounded services. In our case this leads to a very long execution time as the API does not respond very fast when there are more than 100 services bound to the cluster (actually it even results in timeout exceptions).

As we do not even need the information about all bounded services I would like to introduce the possibility to add an additional variable to filter for specific namespaces.

This should improve the overall performance of the data source and also the reliability.
The default value still remains as an empty string, so the changes should be backwards compatible.

In case of any questions, don't hesitate to contact me.

### DIFF
--- a/ibm/data_source_ibm_container_cluster.go
+++ b/ibm/data_source_ibm_container_cluster.go
@@ -284,6 +284,13 @@ func dataSourceIBMContainerCluster() *schema.Resource {
 				Computed: true,
 			},
 
+			"list_bounded_services": {
+				Type:        schema.TypeBool,
+				Default:     true,
+				Optional:    true,
+				Description: "If set to false bounded services won't be listed.",
+			},
+
 			ResourceControllerURL: {
 				Type:        schema.TypeString,
 				Computed:    true,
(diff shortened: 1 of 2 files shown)


ibm/data_source_ibm_container_cluster.go
```diff
@@ -352,7 +359,8 @@ func dataSourceIBMContainerClusterRead(d *schema.ResourceData, meta interface{})
 	for i, worker := range workerFields {
 		workers[i] = worker.ID
 	}
-	servicesBoundToCluster, err := csAPI.ListServicesBoundToCluster(name, "", targetEnv)
+	boundedServicesNamespaceID := d.Get("bounded_services_namespace_id").(string)
```
hkantare, line 362: Since its an optional fields use d.GetOk
```
var namespaceID string
if nID, ok := d.GetOk("bounded_services_namespace_id"); ok {
namespaceID = nID.(string)
}
	servicesBoundToCluster, err := csAPI.ListServicesBoundToCluster(name, namespaceID, targetEnv)
	if err != nil {
		return fmt.Errorf("Error retrieving services bound to cluster: %s", err)
	}
```

hkantare:
Add the attribute to documentaion files also under https://github.com/IBM-Cloud/terraform-provider-ibm/blob/master/website/docs/d/container_cluster.html.markdown

jfeddern:
Hi @hkantare , 
thanks for your feedback.
In the meantime I was thinking about a better solution. As already mentioned in our case its not even needed to fetch any bounded services. Therefore I prefer to introduce a flag as an input field. If the flag is set to true, we won't even use the API in this case.
This would also reduce the overall load for the API endpoint itself. What do you think, does this solution makes more sense?

Kind regards, 
Jan

hkantare:
Hi
Yes flag should be best to control the behaviour. Name the attribute as 
list_bounded_services by default set to true..If flag is set to false you can skip the call to the API...

jfeddern:
Alright, thank you.
I updated the name of the flag, the default value and also the documentation.

PR #2090, opened 2020-12-03
fix: Create new CisRLClient for thread safety

mkistler:
This PR creates a new client struct for each invocation of `CisRLClientSession`. which should eliminate conflicts between different instances of `ibm_cis_rate_limit` that specify different domains.

We will work on adding a new method to the generated Go SDK to make this code more concise in the future, but for now this should fix the problem.


KumarGanesanIBM:
The fix is part of the PR https://github.com/IBM-Cloud/terraform-provider-ibm/pull/2091.  Hence closing this PR.

PR #2110, opened 2020-12-11
Add support for Atracker service

mkistler:
This PR adds support for the ATracker service.

Much of this code was generated by new support for Terraform generation in the `openapi-sdkgen` project.

The commits are intentionally structured as follows:
- the first commit contains the code straight out of the generator
- the second commit is the changes to the provider code to integrate in the new function
- the third commit are hand-edits to fix minor problems in the generated code.

The code currently uses a non-release version of the ATracker SDK, so we should not merge this PR as is.  But once the SDK is updated, we can remove the patch to point to the pre-production SDK and then it should be ready to go.


hkantare:
closing this PR we have another PR support activity tracker

PR #2111, opened 2020-12-12
CIS DNS records import and export development

KumarGanesanIBM:


ibm/resource_ibm_cis_dns_records_import.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+
+	crn := d.Get(cisID).(string)
+	zoneID, _, err := convertTftoCisTwoVar(d.Get(cisDomainID).(string))
+	cisClient.Crn = core.StringPtr(crn)
+	cisClient.ZoneIdentifier = core.StringPtr(zoneID)
+	file := d.Get(cisDNSRecordsImportFile).(string)
+
+	f, err := os.Open(file)
```
kavya498, line 71: Instead of opening file inside resource.. 
Can it be file = `file(<filepath>)` in terraform

ibm/resource_ibm_cis_dns_records_import.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+	opt := cisClient.NewPostDnsRecordsBulkOptions()
+	opt.SetFile(f)
+	result, response, err := cisClient.PostDnsRecordsBulk(opt)
+	if err != nil {
+		log.Printf("Error importing dns records: %v", response)
+		return err
+	}
+	id := fmt.Sprintf("%v:%v:%s:%s:%s", *result.Result.TotalRecordsParsed,
```
kavya498, line 82: not advisable to set id like this..
import scenerio ll fail.. if id has complex structures

ibm/resource_ibm_cis_dns_records_import_test.go
```diff
@@ -0,0 +1,97 @@
...
+	return testAccCheckIBMCisDomainDataSourceConfigBasic1() + fmt.Sprintf(`
+	resource "ibm_cis_dns_records_import" "test" {
+		cis_id    = data.ibm_cis.cis.id
+		domain_id = data.ibm_cis_domain.cis_domain.domain_id
+		file      = "%[1]s"
+	}
+	`, file)
+}
+
+func testAccCheckIBMCisDNSRecordsImportRemoveImportedRecords(n string) resource.TestCheckFunc {
```
kavya498, line 54: if it is been destroyed here.. why is it not added in resource?

Issue #2151, opened 2021-01-08
github.com/IBM-Cloud/terraform-provider-ibm/ibm.expireRuleGet -> panic: runtime error: invalid memory address or nil pointer dereference


mschenk42:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.12.29
+ provider.ibm v1.19.0
+ provider.null v2.1.2

### Affected Resource(s)
- ibm_cos_bucket

```
 panic: runtime error: invalid memory address or nil pointer dereference
 [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2b4c2ff]
 
 goroutine 223 [running]:
 github.com/IBM-Cloud/terraform-provider-ibm/ibm.expireRuleGet(0xc0002d8200, 0x1, 0x1, 0xc0006af900, 0x0, 0x1)
 	/Users/harinireddy/go/src/github.com/IBM-Cloud/terraform-provider-ibm/ibm/structures.go:725 +0x19f
 github.com/IBM-Cloud/terraform-provider-ibm/ibm.resourceIBMCOSRead(0xc0001a7570, 0x3213480, 0xc000289100, 0xc0001a7570, 0x0)
 	/Users/harinireddy/go/src/github.com/IBM-Cloud/terraform-provider-ibm/ibm/resource_ibm_cos_bucket.go:666 +0x112c
 github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).RefreshWithoutUpgrade(0xc000993d80, 0xc000831400, 0x3213480, 0xc000289100, 0xc0004870f0, 0x0, 0x0)
 	/Users/harinireddy/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.6.0/helper/schema/resource.go:455 +0x129
 github.com/hashicorp/terraform-plugin-sdk/internal/helper/plugin.(*GRPCProviderServer).ReadResource(0xc00000e928, 0x36362a0, 0xc00067d9e0, 0xc000831310, 0xc00000e928, 0xc00067d9e0, 0xc0010d7b78)
 	/Users/harinireddy/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk@v1.6.0/internal/helper/plugin/grpc_provider.go:525 +0x3dd
[...]

PR #2156, opened 2021-01-12
invalid memory address or nil pointer dereference ibm.expireRuleGet

mschenk42:
Issue #2151

### DIFF
--- a/ibm/structures.go
+++ b/ibm/structures.go
@@ -722,7 +722,7 @@ func expireRuleGet(in []*s3.LifecycleRule) []interface{} {
 				rule["days"] = int(*(r.Expiration).Days)
 			}
 			if r.Filter != nil {
-				rule["prefix"] = *(r.Filter).Prefix
+				rule["prefix"] = r.Filter.Prefix
 			}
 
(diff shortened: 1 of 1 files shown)


ibm/structures.go
```diff
@@ -722,7 +722,7 @@ func expireRuleGet(in []*s3.LifecycleRule) []interface{} {
 				rule["days"] = int(*(r.Expiration).Days)
 			}
 			if r.Filter != nil {
-				rule["prefix"] = *(r.Filter).Prefix
+				rule["prefix"] = r.Filter.Prefix
```
hkantare, line 725: This should n't be this way to handle the nil ...If we delete * then it will stores the value of the address
We have PR from COS team who fixed it to handle it
https://github.com/IBM-Cloud/terraform-provider-ibm/pull/2159

mschenk42: ah, ok, I'll test out this other fix #2159  locally and report back here

mschenk42: @hkantare I tested #2159 and that worked without issue.  When do you plan to cut a new release with this fix? Thanks.

PR #2162, opened 2021-01-13
Event Streams fix: setId to be empty when topic does not exist

JunliWang:


ibm/resource_ibm_event_streams_topic.go
```diff
@@ -183,6 +182,7 @@ func resourceIBMEventStreamsTopicRead(d *schema.ResourceData, meta interface{})
 			return nil
 		}
 	}
+	d.SetId("")
```
hkantare, line 185: I think just setting the ID is not enough because even after setting the the ID here we are returing error and failing out. We should set the ID to empty and return successfully

Can you please test this scenario to validate this

1)create topic via terraform
2)Delete the topic manually uisng ui or ibmcloud cli
3)Run terraform plan (Here it shd plan out saying 1 new resource to be added)

hkantare:
Can you please rebase and add the fix because we can see old commits also part of the pr

hkantare:
@JunliWang can you please check on review comments

JunliWang:
closed this PR and created a new one: https://github.com/IBM-Cloud/terraform-provider-ibm/pull/2220

Issue #2058, opened 2020-11-21
Data source 'ibm_is_subnet' causes Terraform crash when subnet not found


jaywcarman:
### Terraform Version
Terraform v0.12.20
+ provider.ibm v1.14.0

### Affected Resource(s)
- ibm_is_subnet

### Terraform Configuration Files
```hcl
provider "ibm" {
    version          = ">= 1.14.0"
    generation       = "2"
    region           = "us-south"
}
data ibm_is_subnet "thissubnetdoesnotexist" {
  name = "this_subnet_does_not_exist"
}
```

### Debug Output
https://gist.github.com/jaywcarman/5219f64e5248bfcc2d2953e1cfe1e783

### Panic Output
https://gist.github.com/jaywcarman/5b94252823caa3a25a4c7957b822dd30

PR #2171, opened 2021-01-18
issue 2058 update data_source_ibm_is_subnet.go

deepaksibm:


deepaksibm:
Closes #2058

ibm/data_source_ibm_is_subnet.go
```diff
@@ -238,6 +238,10 @@ func subnetGetByNameOrID(d *schema.ResourceData, meta interface{}) error {
 				break
 			}
 		}
+		if subnet == nil {
```
hkantare, line 241: Can you change the error message something like below

return fmt.Errorf("No subnet with name [%s]", name)

PR #2183, opened 2021-01-20
fixed keyname lookup issue#2158

VaishnaviGopal:


ibm/resource_ibm_storage_file.go
```diff
@@ -1230,7 +1230,7 @@ func getSaaSPerformSpacePrice(productItems []datatypes.Product_Item, size int) (
 		}
 
 		keyname := fmt.Sprintf("%d_%d_GBS", capacityMinimum, capacityMaximum)
-		if *item.KeyName != keyname {
+		if !strings.Contains(*item.KeyName, keyname) {
```
hkantare, line 1233: please add the condition to check item.keyname first not nil and then other condition

if item.KeyName == nil || !strings.Contains(*item.KeyName, keyName) {

ibm/resource_ibm_storage_file.go
```diff
@@ -1291,7 +1291,7 @@ func getSaaSEnduranceSpacePrice(productItems []datatypes.Product_Item, size int,
 
 	for _, item := range productItems {
 
-		if *item.KeyName != keyName {
```
hkantare, line 1294: same as above

Issue #1196, opened 2020-03-18
support for pod-subnet and service-subnet


krakel:
At the command line ibmcloud support the --pod-subnet and --service-subnet option.

https://www.ibm.com/cloud/blog/using-custom-subnet-ranges-with-ibm-cloud-kubernetes-service
https://cloud.ibm.com/docs/containers?topic=containers-cli-plugin-kubernetes-service-cli#pod-subnet

Does this option also exist in the cloud provider?

gregswift:
Do we have any idea what this would take to implement? Its a requirement for a cluster we were getting ready to deploy with terraform.

bradbehle:
@gregswift I talked to @hkantare via slack (hkantare@in.ibm.com) and he said they would work on this in their next sprint

hkantare:
Released in https://github.com/IBM-Cloud/terraform-provider-ibm/releases/tag/v1.20.0
Cloud docs will be updated soon
Registry docs:
https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs

kavya498:
Closing this issue.

PR #2190, opened 2021-01-21
Support #1196: Pod Subnet and Service Subnet for classic clusters

kavya498:


gregswift:
@hkantare thank you so much! any idea when yall are cutting the next release?

PR #2286, opened 2021-02-26
Enhancement : added default routing table id to output parameters of ibm_is_vpc

ujjwal-ibm:
Added an output parameter "default_routing_table" which returns the id of the default routing table created


ibm/resource_ibm_is_vpc.go
```diff
@@ -89,6 +90,14 @@ func resourceIBMISVPC() *schema.Resource {
 				Description: "Default network ACL",
 			},
 
+			isVPCDefaultRoutingTable: {
+				Type:        schema.TypeString,
```
hkantare, line 94: isVPCDefaultRoutingTable here should be only computed  and we need to set in Read Method similar to datasource

PR #2293, opened 2021-03-01
Key alias

harshit777:


ibm/resource_ibm_kms_key_alias.go
```diff
@@ -0,0 +1,248 @@
...
+		Importer: &schema.ResourceImporter{},
+
+		Schema: map[string]*schema.Schema{
+			"instance_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Key ID",
+				ForceNew:    true,
+			},
+			"alias_name": {
```
hkantare, line 28: Do we have alias_name and alias_id bith from API..if not just rename this as alias

ibm/data_source_ibm_kms_key.go
```diff
@@ -22,9 +22,16 @@ func dataSourceIBMKMSkey() *schema.Resource {
...
 			"key_name": {
-				Type:        schema.TypeString,
-				Required:    true,
-				Description: "The name of the key to be fetched",
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "The name of the key to be fetched",
+				ExactlyOneOf: []string{"alias_name", "key_name"},
+			},
+			"alias_name": {
```
hkantare, line 30: here also similar just use alias if we don't have name and id

ibm/data_source_ibm_kms_key.go
```diff
@@ -192,40 +204,72 @@ func dataSourceIBMKMSKeyRead(d *schema.ResourceData, meta interface{}) error {
...
-	}
-	var keyName string
-	var matchKeys []kp.Key
-	if v, ok := d.GetOk("key_name"); ok {
-		keyName = v.(string)
-		for _, keyData := range retreivedKeys {
-			if keyData.Name == keyName {
-				matchKeys = append(matchKeys, keyData)
+
+	if d.Get("alias_name").(string) == "" {
```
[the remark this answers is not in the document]
harshit777: I am checking the same inside the alias `if condition` if the user has not used alias_name then it checks for d.GetOk("Key_name") line 220.

PR #2296, opened 2021-03-02
Enhancement Trusted image checksum

deepaksibm:


ibm/structures.go
```diff
@@ -51,6 +51,8 @@ const (
 	ResourceName = "resource_name"
 	//ResourceCRN ...
 	ResourceCRN = "resource_crn"
+	//ResourceCheckSum
+	ResourceCheckSum = "resource_checksum"
```
hkantare, line 55: don't add ResourceCheckSum here...it should be part of the resource/datasource..

The ResourceName,ResourceCRN  ...are some generic arguments added for multiple resources

ibm/data_source_ibm_is_images.go
```diff
@@ -180,6 +185,9 @@ func imageList(d *schema.ResourceData, meta interface{}) error {
 			"os":           *image.OperatingSystem.Name,
 			"architecture": *image.OperatingSystem.Architecture,
 		}
+		if image.File != nil && image.File.Checksums != nil && image.File.Checksums.Sha256 != nil {
+			l["checksum"] = *image.File.Checksums.Sha256
```
hkantare, line 189: image.File.Checksums.Sha256 != nil this is not required since  image.File.Checksums.Sha256 is a string not a complex structure

hkantare: Add respective documentations files datasource/resource with new attribute

PR #2332, opened 2021-03-10
Logging not supported for Network Loadbalancer

SunithaGudisagarIBM:
Screenshot 2021-03-10 at 2 12 47 PM

Screenshot 2021-03-10 at 2 20 56 PM
Screenshot 2021-03-10 at 2 29 54 PM
Screenshot 2021-03-10 at 2 52 58 PM

### DIFF
--- a/ibm/resource_ibm_is_lb.go
+++ b/ibm/resource_ibm_is_lb.go
@@ -108,12 +108,13 @@ func resourceIBMISLB() *schema.Resource {
 			},
 
 			isLBProfile: {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Computed:     true,
-				ForceNew:     true,
-				Description:  "The profile to use for this load balancer.",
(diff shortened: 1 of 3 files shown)


ibm/resource_ibm_is_lb.go
```diff
@@ -501,6 +508,16 @@ func lbGet(d *schema.ResourceData, meta interface{}, id string) error {
 		if profile.Name != nil {
 			d.Set(isLBProfile, *lb.Profile.Name)
 		}
+	} else {
+		if lb.Logging != nil {
```
hkantare, line 512: here you can have only 1 if condition if lb.logging != nil && lb.logging.Datapath != nil &&l b.logging.Datapath.Active != nil

SunithaGudisagarIBM: @hkantare Changes Incorporated

PR #2353, opened 2021-03-15
enhancement: added tag support for is_security_group

ujjwal-ibm:
Added tagging support for security group

### DIFF
--- a/ibm/data_source_ibm_is_security_group.go
+++ b/ibm/data_source_ibm_is_security_group.go
@@ -4,6 +4,7 @@
 package ibm
 
 import (
+	"log"
 	"reflect"
 
 	"github.com/IBM/vpc-go-sdk/vpcclassicv1"
@@ -24,6 +25,8 @@ const (
 	isSgRulePortMin   = "port_min"
 	isSgRuleProtocol  = "protocol"
 	isSgVPC           = "vpc"
+	isSgTags          = "tags"
+	isSgCRN           = "crn"
 )
 
 func dataSourceIBMISSecurityGroup() *schema.Resource {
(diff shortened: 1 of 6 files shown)


ibm/data_source_ibm_is_security_group.go
```diff
@@ -128,6 +131,21 @@ func dataSourceIBMISSecurityGroup() *schema.Resource {
 				Computed:    true,
 				Description: "The resource group name in which resource is provisioned",
 			},
+
+			isSgTags: {
+				Type:        schema.TypeSet,
+				Optional:    true,
```
hkantare, line 137: in datasource tags it only computed variable

ibm/resource_ibm_is_security_group.go
```diff
@@ -472,6 +530,14 @@ func resourceIBMISSecurityGroupUpdate(d *schema.ResourceData, meta interface{})
 	} else {
 		return resourceIBMISSecurityGroupRead(d, meta)
 	}
+	if d.HasChange(isSecurityGroupTags) {
```
hkantare, line 533: The logic of code needs to be changed here

```
if d.HasChange(isSecurityGroupName) {
		name = d.Get(isSecurityGroupName).(string)
		hasChanged = true
	} 
	if d.HasChange(isSecurityGroupTags) {
		oldList, newList := d.GetChange(isSecurityGroupTags)
		err = UpdateTagsUsingCRN(oldList, newList, meta, d.Get(isSecurityGroupCRN).(string))
		if err != nil {
			log.Printf(
				"Error Updating Security Group tags: %s\n%s", d.Id(), err)
		}
	}
       if hasChanged {
	if userDetails.generation == 1 {
		err := classicSgUpdate(d, meta, id, name, hasChanged)
		if err != nil {
			return err
		}
	} else {
		err := sgUpdate(d, meta, id, name, hasChanged)
		if err != nil {
			return err
		}
	}
	}
	return resourceIBMISSecurityGroupRead(d, meta)
	```

hkantare: Try to test the scenario wher u up[date name and tags at once

PR #2370, opened 2021-03-18
Fix the default RC endpoint

mengqxqx:
Default endpoint should not include `/v2`


kavya498:
@mengqxqx ,
SDK also include /v2 for default endpoint.. I believe it doesn't work without /v2

mengqxqx:
@kavya498 
Update the hard code URL to use RM DefaultServiceURL.

For remove `/v2` or not, @phil_adams@us.ibm.com is looking into it.

PR #2372, opened 2021-03-18
Replace container registry provider with generated variant

jhart1685:
Which also introduces a retention policy resource type.

Attn of @hkantare @mkistler 
Note - I had some issues with the detect-secrets pre-commit hook. It looks like running this from "master" runs `ibm.29.dss` version, regardless of whether you have a later version (30 or 31) of the developer tool locally. Hopefully nothing gets broken by any modifications I've made to .secrets.baseline

The bulk of the code, tests, examples and docs for this were generated by Mike.
I've added `created_on` / `updated_on` as deprecated fields, to contain the same values as `created_date` / `updated_date`.
Also noticed that the data source was missing some fields, so added those.


ibm/resource_ibm_cr_retention_policy.go
```diff
@@ -0,0 +1,183 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"namespace": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The namespace to which the retention policy is attached.",
+			},
+			"images_per_repo": &schema.Schema{
+				Type:        schema.TypeInt,
+				Optional:    true,
```
hkantare, line 46: Based on this comment
// A retention policy "does not exist" if `imagesPerRepo` and `retainUntagged` have their default values

Since you are definign default values  for images_per_repo and retain_untagged then if user provides
resource "ibm_cr_retention_policy" "cr_retention_policy" {
			namespace = ibm_cr_namespace.cr_namespace.name
		}
		
		Then will API throws back an error imagesPerRepo and retainUntagged both are set to default vlaues?
		or can we add condition check and throws back an error befor calling the create retention policy?

website/docs/r/cr_namespace.html.markdown
```diff
@@ -1,61 +1,39 @@
 ---
-
-subcategory: "Container Registry"
```
hkantare, line 3: please include this subcategory and remove sidebar_current

This changes we added recently for Terraform hashicorp registry docs

website/docs/r/cr_namespace.html.markdown
```diff
@@ -1,61 +1,39 @@
...
-```
-$ terraform import ibm_cr_namespace.test <name of the namespace>
-
-$ terraform import ibm_cr_namespace.test namespace-name
-```
+* `id` - The unique identifier of the cr_namespace.
+* `account` - The IBM Cloud account that owns the namespace.
+* `created_date` - When the namespace was created.
+* `crn` - If the namespace has been assigned to a resource group, this is the IBM Cloud CRN representing the namespace.
+* `resource_created_date` - When the namespace was assigned to a resource group.
```
hkantare, line 38: Please include the terraform import section also back

website/docs/r/cr_retention_policy.html.markdown
```diff
@@ -0,0 +1,35 @@
+---
+layout: "ibm"
```
hkantare, line 2: Add subcategory and remove sidebar_current

website/docs/r/cr_retention_policy.html.markdown
```diff
@@ -0,0 +1,35 @@
...
+
+* `namespace` - (Required, string) The namespace to which the retention policy is attached.
+* `images_per_repo` - (Optional, int) Determines how many images will be retained for each repository when the retention policy is executed. The value -1 denotes 'Unlimited' (all images are retained).
+* `retain_untagged` - (Optional, bool) Determines if untagged images are retained when executing the retention policy. This is false by default meaning untagged images will be deleted when the policy is executed.
+
+## Attribute Reference
+
+The following attributes are exported:
+
+* `id` - The unique identifier of the cr_retention_policy.
```
hkantare, line 35: Add the import section...This will helps users how can he bring back the infrastructure which he created using ui, api,cli ...to terraform state management

ibm/data_source_ibm_cr_namespaces.go
```diff
@@ -1,19 +1,36 @@
-// Copyright IBM Corp. 2017, 2021 All Rights Reserved.
-// Licensed under the Mozilla Public License v2.0
+/**
+ * (C) Copyright IBM Corp. 2021.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
mkistler: The generator has been updated and I'll regenerate to pick up this change.

ibm/resource_ibm_cr_namespace.go
```diff
@@ -1,214 +1,234 @@
...
 	if err != nil {
-		return err
+		log.Printf("[DEBUG] CreateNamespaceWithContext failed %s\n%s", err, response)
+		return diag.FromErr(err)
 	}
-	d.SetId(response.Namespace)
-	return resourceIBMContainerRegistryNamespaceRead(d, meta)
+
+	d.SetId(*namespace.Namespace)
+
```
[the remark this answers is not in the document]
mkistler: @hkantare Can you please open an issue on the generator for this feature?  Currently the generator only supports "local tags", but I agree that we should use ghost where appropriate.

ibm/resource_ibm_cr_namespace_test.go
```diff
@@ -11,125 +24,158 @@ import (
...
+	"github.com/IBM/container-registry-go-sdk/containerregistryv1"
 )
 
 func TestAccIBMCrNamespaceBasic(t *testing.T) {
-
-	namespaceName := fmt.Sprintf("terraform-tf-%d", acctest.RandIntRange(10, 100))
-	namespaceName1 := fmt.Sprintf("terraform-tf-%d", acctest.RandIntRange(10, 100))
-	resourceName := "ibm_cr_namespace.test_namespace"
+	var conf containerregistryv1.NamespaceDetails
+	name := fmt.Sprintf("name_%d", acctest.RandIntRange(10, 100))
```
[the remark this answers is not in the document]
mkistler: I think this is a change we should make in the generator.  Can you please open an issue for this?

ibm/resource_ibm_cr_retention_policy.go
```diff
@@ -0,0 +1,183 @@
...
+
+	setRetentionPolicyOptions.SetNamespace(d.Get("namespace").(string))
+	if _, ok := d.GetOk("images_per_repo"); ok {
+		setRetentionPolicyOptions.SetImagesPerRepo(int64(d.Get("images_per_repo").(int)))
+	}
+	if _, ok := d.GetOk("retain_untagged"); ok {
+		setRetentionPolicyOptions.SetRetainUntagged(d.Get("retain_untagged").(bool))
+	}
+
+	response, err := containerRegistryClient.SetRetentionPolicyWithContext(context, setRetentionPolicyOptions)
```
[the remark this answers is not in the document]
mkistler: The accountID is a "global parameter" for the SDK and is set as part of client initialization:

https://github.com/IBM-Cloud/terraform-provider-ibm/pull/2372/files#diff-6dc66490d06e928e5da63fa4d3ce2c6410a6d6bdacb2087026546d583b8c5527R1150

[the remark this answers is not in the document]
jhart1685: Also , it's not needed in the payload here (ignored if set)..

website/docs/r/cr_namespace.html.markdown
mkistler: We'll update the generator to address this.

website/docs/r/cr_namespace.html.markdown
mkistler: We'll update the generator to address this.

ibm/resource_ibm_cr_namespace.go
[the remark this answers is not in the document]
mkistler: Sadly, this turns out to be difficult for namespaces, since the response to a namespace create _does not_ provide the CRN of the namespace.  So before I spend time working out the hand-edits that are needed to make this work for namespaces, I want to ask how important it is to have "formal" tags for namespaces instead of "local" tags.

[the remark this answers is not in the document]
mkistler: I will further note that the UI for namespaces does not show tags associated with a namespace (as best I can tell), and namespaces do not show up in the "Resource list", so I don't see any way to verify that tags were applied to the namespace if indeed we can figure out how to do it.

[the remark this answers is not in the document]
jhart1685: Agree with Mike here.. There is further context (but for a different forum).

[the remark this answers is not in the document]
hkantare: Ok I agree with Mike and James...you can for now ignore tags

jhart1685:
@hkantare - most of the comments are addressed by the latest generator (3.30).
But not quite all. In particular, there are a couple of your points around the html doc that will benefit from a little more work in the generator before we resuscitate this PR. Hopefully not too much longer.

ibm/resource_ibm_cr_namespace.go
```diff
@@ -1,214 +1,234 @@
...
+	for _, namespaceDetails = range namespaceDetailsList {
+		if *namespaceDetails.Name == d.Id() {
+			break
+		}
 	}
-	namespace := d.Id()
-	target := registryv1.NamespaceTargetHeader{
-		AccountID: accountID,
+	if namespaceDetails.Name == nil || *namespaceDetails.Name != d.Id() {
+		return diag.FromErr(fmt.Errorf("Namespace %s not found", d.Id()))
```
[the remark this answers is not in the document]
jhart1685: Think this is now done

website/docs/r/cr_namespace.html.markdown
jhart1685: Done

website/docs/r/cr_namespace.html.markdown
jhart1685: Done

website/docs/r/cr_retention_policy.html.markdown
jhart1685: Done

website/docs/r/cr_retention_policy.html.markdown
jhart1685: Think this is now done

ibm/resource_ibm_cr_namespace_test.go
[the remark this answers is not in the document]
jhart1685: Done

ibm/data_source_ibm_cr_namespaces.go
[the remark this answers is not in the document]
jhart1685: Done

ibm/resource_ibm_cr_retention_policy.go
jhart1685: Thinking...

jhart1685: Ok - firstly, I will correct that comment... a retention policy does not exist if `imagesPerRepo` is `-1` and `retainUntagged` is `true` , the default value for `retainUntagged` when creating or updating a policy is actually `false`.
However, I think the code is mostly accurate in this regard.

I agree it would make sense, in the context of creating a retention policy, to throw back an error if images_per_repo is not set. The equivalent CLI command for this treats this parameter as mandatory. 
I'll fix this up.

jhart1685: Believe I have addressed this on line 58-60:
```Go
} else {
		return diag.Errorf("images_per_repo must be set when creating a retention policy")
	}
```

jhart1685:
Hi @hkantare - I think I've addressed the comments.
As per separate discussions, I also hit a gpg signing key issue when attempting to test locally (and in a Jenkins job). I found that updating the terraform-plugin-sdk dependency to v2.6.1 resolved that, so is included in this PR - but obviously am unsure whether you want to make that leap.
I also don't know what the "1 workflow awaiting approval" warning means here. As a contributor from a fork, I'm surprised that I need to run workflows.

ibm/resource_ibm_cr_retention_policy.go
jhart1685: Hmm - maybe this isn't right. In local testing, it looks like Terraform just uses the default value in this case (-1), and the `else` block doesn't get hit.
@hkantare - Do you have any guidance here? Should we check for it being set to -1 here, instead of (or as well as) checking for it not being set at all?

jhart1685: ok - I've refixed this as outlined in my previous comment. This now works when tested locally.

ibm/data_source_ibm_cr_namespaces.go
```diff
@@ -54,38 +81,42 @@ func dataIBMContainerRegistryNamespaces() *schema.Resource {
...
 	if err != nil {
-		return err
-	}
-	target := registryv1.NamespaceTargetHeader{
-		AccountID: accountID,
+		return diag.FromErr(err)
 	}
 
-	crAPI := crClient.Namespaces()
+	listNamespaceDetailsOptions := &containerregistryv1.ListNamespaceDetailsOptions{}
```
hkantare, line 90: The code releated/ to set the accoutnID is removed from options...Will this effect by any chance if user uses their serviceID API key

ibm/data_source_ibm_cr_namespaces.go
```diff
@@ -54,38 +81,42 @@ func dataIBMContainerRegistryNamespaces() *schema.Resource {
...
 
-	response, err := crAPI.GetDetailedNamespaces(target)
+	namespaceDetailsList, response, err := containerRegistryClient.ListNamespaceDetails(listNamespaceDetailsOptions)
 	if err != nil {
-		return err
+		if response != nil && response.StatusCode == 404 {
```
hkantare, line 94: in datasource don't addd code to check if response != nil && response.StatusCode == 404 { ----}
just return back the err

ibm/resource_ibm_cr_retention_policy.go
hkantare:
```
I agree it would make sense, in the context of creating a retention policy, to throw back an error if images_per_repo is not set. The equivalent CLI command for this treats this parameter as mandatory.
I'll fix this up.
```

Then can terraform also make  images_per_repo as required because still i'm in confusion if user writes a template
```
resource "ibm_cr_retention_policy" "cr_retention_policy" {
namespace = ibm_cr_namespace.cr_namespace.name
}
```

Then by default images_per_repo is -1 and terraform apply fails
````
images_per_repo must be set when creating a retention policy
```
That implies that images_per_repo is required parameter

ibm/data_source_ibm_cr_namespaces.go
mkistler: The AccountID is a "global parameter" -- it is set in [`ibm/config.go`](https://github.com/IBM-Cloud/terraform-provider-ibm/pull/2372/files#diff-6dc66490d06e928e5da63fa4d3ce2c6410a6d6bdacb2087026546d583b8c5527R1280) as part of the initialization of the Container Registry SDK.

jhart1685: What Mike said....

ibm/data_source_ibm_cr_namespaces.go
jhart1685: Hmm - I think this was one of Mike's hand-edits. This is the case where the account doesn't have any namespaces in that region.. @mkistler - I'm trying to recall why this was needed?

ibm/resource_ibm_cr_retention_policy.go
jhart1685: @hkantare - yes, we could go down that route... 
Strictly speaking, you might want to update a retention policy to change the retain_untagged behaviour, without modifying images_per_repo.. The API allows this, but our CLI won't currently let you do it - so probably reasonable just to make it a required parameter. I'll make the change, and let you know when it's ready for re-review.

To answer your question directly - if you were to create this policy (if our provider allowed it):
```
resource "ibm_cr_retention_policy" "cr_retention_policy" {
namespace = ibm_cr_namespace.cr_namespace.name
}
```

You would end up with all images retained, but the retain_untagged would flip from true to false.. This is also a bit esoteric and not easy for users to understand - which strengthens the argument for just making images_per_repo mandatory.

jhart1685:
@hkantare - I've now made images_per_repo Required.

ibm/data_source_ibm_cr_namespaces.go
jhart1685: I checked with Mike - and he agrees with you. I made the change and have tested it.

PR #2381, opened 2021-03-22
feature: bump bluemix-go version to latest

mindjiver:
To support the new visibility feature of upstream bluemix-go bump the
version.


kavya498:
We have few more changes to be done on bluemix-go..
Once it is done we ll include..

mindjiver:
Cool @kavya498 , we just needed this for a MZR that was "removed" and bumping this library did the trick. We are all fine with waiting if you are doing more improvement in the bluemix-go library.

hkantare:
closing this we already vendored latest now

PR #2400, opened 2021-03-24
Migrate the resource-key resource to Platform-go-sdk

mengqxqx:


ibm/resource_ibm_resource_key.go
```diff
@@ -97,12 +96,96 @@ func resourceIBMResourceKey() *schema.Resource {
 				Computed:    true,
 				Description: "crn of resource key",
 			},
+
```
kavya498, line 99: All these attributes have to be set in READ Method

ibm/resource_ibm_resource_key.go
```diff
@@ -184,28 +263,32 @@ func resourceIBMResourceKeyUpdate(d *schema.ResourceData, meta interface{}) erro
...
+	}
 
-	resourceKey, err := rsContClient.ResourceServiceKey().GetKey(resourceKeyID)
+	resourceKey, resp, err := rsContClient.GetResourceKey(&resourceKeyGet)
 	if err != nil {
-		return fmt.Errorf("Error retrieving resource key: %s", err)
+		return fmt.Errorf("Error retrieving resource key: %s with resp code: %s", err, resp)
 	}
-	d.Set("credentials", Flatten(resourceKey.Credentials))
+	d.Set("credentials", resourceKey.Credentials)
```
kavya498, line 279: It would break existing customer..Schema is of type map[string]interface{}
have to conver to map[string]interface{} and set.. something like..
```

var credInterface map[string]interface{}
cred, _ := json.Marshal(resourceKey.Credentials)
json.Unmarshal(cred, &credInterface)
d.Set("credentials", Flatten(credInterface))
```

ibm/resource_ibm_resource_key.go
```diff
@@ -184,28 +263,32 @@ func resourceIBMResourceKeyUpdate(d *schema.ResourceData, meta interface{}) erro
...
 	d.Set("name", resourceKey.Name)
 	d.Set("status", resourceKey.State)
-	if roleCrn, ok := resourceKey.Credentials["iam_role_crn"].(string); ok {
-
+	iamRoleCrn := resourceKey.Credentials.IamRoleCRN
+	if *iamRoleCrn != "" {
+		roleCrn := *iamRoleCrn
 		roleName := roleCrn[strings.LastIndex(roleCrn, ":")+1:]
 
-		if strings.Contains(roleCrn, ":customRole:") {
```
kavya498, line 204: Revert back to customRole..
This is the generic resource and this piece of code is to set back custom roles

ibm/resource_ibm_resource_key.go
```diff
@@ -184,28 +263,32 @@ func resourceIBMResourceKeyUpdate(d *schema.ResourceData, meta interface{}) erro
...
+		if strings.Contains(roleCrn, ":serviceRole:") {
 			iampapv2Client, err := meta.(ClientSession).IAMPAPAPIV2()
 			if err == nil {
-				resourceCRN := resourceKey.Crn.ServiceName
-				roles, err := iampapv2Client.IAMRoles().ListCustomRoles(resourceKey.AccountID, resourceCRN)
+				resourceCRN := resourceKey.CRN
```
kavya498, line 290: Service Name is used to list custom roles..
Earlier CRN was object but here it is a string..so have to obtain service name from the CRN string
```
var resourceCRN string
if instance.CRN != nil {
		serviceName := strings.Split(*instance.CRN, ":")
		if len(serviceName) > 4 {
			resourceCRN =serviceName[4]
		}
	}
```

ibm/resource_ibm_resource_key.go
```diff
@@ -97,12 +96,96 @@ func resourceIBMResourceKey() *schema.Resource {
 				Computed:    true,
 				Description: "crn of resource key",
 			},
+
+			"guid": {
```
kavya498, line 100: All the additional attributes added in schema has to be set  in the READ method

ibm/resource_ibm_resource_key.go
```diff
@@ -5,16 +5,15 @@ package ibm
...
 	"time"
 
+	rc "github.com/IBM/platform-services-go-sdk/resourcecontrollerv2"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
 	"github.com/IBM-Cloud/bluemix-go/api/iampap/iampapv2"
-	"github.com/IBM-Cloud/bluemix-go/api/resource/resourcev1/controller"
-	"github.com/IBM-Cloud/bluemix-go/bmxerror"
-	"github.com/IBM-Cloud/bluemix-go/crn"
 	"github.com/IBM-Cloud/bluemix-go/models"
```
kavya498, line 17: Do bluemix-go dependency still exist?

ibm/resource_ibm_resource_key.go
mengqxqx: ok

ibm/resource_ibm_resource_key.go
mengqxqx: done

ibm/resource_ibm_resource_key.go
mengqxqx: done

ibm/resource_ibm_resource_key.go
mengqxqx: done

PR #2406, opened 2021-03-24
Add Terraform support for ATracker service

mkistler:
This PR adds the Terraform support for the new ATracker service.

The code is mostly generated from our Terraform generator, with just a few small hand-edits.

The new code has not been tested as I do not have access to a test environment.

Marking as Draft for now until testing has been completed.

### DIFF
--- a/ibm/config.go
+++ b/ibm/config.go
@@ -46,6 +46,7 @@ import (
 	cisratelimitv1 "github.com/IBM/networking-go-sdk/zoneratelimitsv1"
 	cisdomainsettingsv1 "github.com/IBM/networking-go-sdk/zonessettingsv1"
 	ciszonesv1 "github.com/IBM/networking-go-sdk/zonesv1"
+	"github.com/IBM/platform-services-go-sdk/atrackerv1"
 	iamidentity "github.com/IBM/platform-services-go-sdk/iamidentityv1"
 	resourcecontroller "github.com/IBM/platform-services-go-sdk/resourcecontrollerv2"
 	resourcemanager "github.com/IBM/platform-services-go-sdk/resourcemanagerv2"
@@ -181,6 +182,7 @@ type ClientSession interface {
 	BluemixAcccountAPI() (accountv2.AccountServiceAPI, error)
 	BluemixAcccountv1API() (accountv1.AccountServiceAPI, error)
 	BluemixUserDetails() (*UserConfig, error)
+	AtrackerV1() (*atrackerv1.AtrackerV1, error)
 	ContainerAPI() (containerv1.ContainerServiceAPI, error)
 	VpcContainerAPI() (containerv2.ContainerServiceAPI, error)
 	ContainerRegistryAPI() (registryv1.RegistryServiceAPI, error)
@@ -257,6 +259,9 @@ type clientSession struct {
 
 	accountV1ConfigErr     error
 	bmxAccountv1ServiceAPI accountv1.AccountServiceAPI
+
+	atrackerClient    *atrackerv1.AtrackerV1
+	atrackerClientErr error
 
 	bmxUserDetails  *UserConfig
 	bmxUserFetchErr error
@@ -494,6 +499,10 @@ func (sess clientSession) BluemixSession() (*bxsession.Session, error) {
 // BluemixUserDetails ...
 func (sess clientSession) BluemixUserDetails() (*UserConfig, error) {
 	return sess.bmxUserDetails, sess.bmxUserFetchErr
+}
+
+func (session clientSession) AtrackerV1() (*atrackerv1.AtrackerV1, error) {
+	return session.atrackerClient, session.atrackerClientErr
 }
 
 // ContainerAPI provides Container Service APIs ...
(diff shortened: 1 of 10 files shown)


ibm/config.go
```diff
@@ -1097,6 +1106,29 @@ func (c *Config) ClientSession() (interface{}, error) {
 		}
 	}
 
+	// Construct an "options" struct for creating the service client.
+	atrackerClientURL, err := atrackerv1.GetServiceURLForRegion("private." + c.Region)
```
hkantare, line 1110: Here by default the atracker is supported by private endpoints. We recently introduced the visibility parameter at provider level. By default its public...So if c.visibility == public may be can we set the error to session.atrackerClientErr  = Public endpoints is  not supported by this service

ibm/data_source_ibm_atracker_routes.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	listRoutesOptions := &atrackerv1.ListRoutesOptions{}
+
+	routeList, response, err := atrackerClient.ListRoutesWithContext(context, listRoutesOptions)
+	if err != nil {
+		log.Printf("[DEBUG] ListRoutesWithContext failed %s\n%s", err, response)
+		return diag.FromErr(err)
```
hkantare, line 94: @mkistler Can we enhance the error to return the error as combination of both err and response...because response contains more details of failure than error?

ibm/data_source_ibm_atracker_routes.go
```diff
@@ -0,0 +1,185 @@
...
+				matchRoutes = append(matchRoutes, data)
+			}
+		}
+	} else {
+		matchRoutes = routeList.Routes
+	}
+	routeList.Routes = matchRoutes
+
+	if len(routeList.Routes) == 0 {
+		return diag.FromErr(fmt.Errorf("no Routes found with name %s\nIf not specified, please specify more filters", name))
```
hkantare, line 116: here no Routes found with name %s\nIf not specified, please specify more filters this error message is confusing...

Here we have only filter name  and again asking to specifiy more filters

ibm/data_source_ibm_atracker_targets_test.go
```diff
@@ -0,0 +1,53 @@
...
+					resource.TestCheckResourceAttr("data.ibm_atracker_targets.atracker_targets", "targets.0.target_type", targetTargetType),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckIBMAtrackerTargetsDataSourceConfigBasic(targetName string, targetTargetType string) string {
+	return fmt.Sprintf(`
+		resource "ibm_atracker_target" "atracker_target" {
```
hkantare, line 38: I think you need to rewrite the testcase to provsion first the dependent resources(cos, buckets...) and then resource "ibm_atracker_target"

or else the prereq here is we should have cos and bucket provsioned and read those variables from provider
_test.go(via enviornmental variables)

ibm/resource_ibm_atracker_route.go
```diff
@@ -0,0 +1,211 @@
...
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(60 * time.Minute),
+			Update: schema.DefaultTimeout(20 * time.Minute),
+			Delete: schema.DefaultTimeout(10 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
```
hkantare, line 34: Can we have a validator added with some regex

ibm/resource_ibm_atracker_target.go
```diff
@@ -0,0 +1,238 @@
...
+							Description: "The CRN of this COS instance.",
+						},
+						"bucket": &schema.Schema{
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "The bucket name under this COS instance.",
+						},
+						"api_key": &schema.Schema{
+							Type:        schema.TypeString,
+							Required:    true,
```
hkantare, line 70: Add sensitive true

ibm/config.go
mkistler: This will be a hand edit.

ibm/data_source_ibm_atracker_routes.go
mkistler: We can.  Can you point to an example in the current IBM Terraform provider that I can mimic?

ibm/data_source_ibm_atracker_targets_test.go
mkistler: I think this will need to be a hand edit.

ibm/data_source_ibm_atracker_routes.go
mkistler: Agreed that this is confusing.  I think we took this message text from an existing data source, perhaps `data_source_ibm_cis.go`, but the situation there is different so it was the wrong pattern to follow.

Further, I don't think we can depend on `name` being set here.

I propose that
1) this check gets moved inside `if suppliedFilter` then block, and
2) the message is changed to simply:
> "no Routes found with name %s", name)

Does this sound good?

ibm/resource_ibm_atracker_route.go
mkistler: The API doc currently specifies no `pattern`, `minLength`, or `maxLength`.  If these are added, an appropriate validator will be generated.

ibm/resource_ibm_atracker_target.go
mkistler: We need to add `x-terraform-sensitive: true` to this property in the API definition.  That will cause the generator to add `Sensitive: true,` here.

ibm/data_source_ibm_atracker_routes.go
mkistler: Like this:
```
return fmt.Errorf("[DEBUG] Create SSH Key %s\n%s", err, response)
```

Or "beautify_error" ...

hkantare:
closing this PR as he had another PR to add Atracker support

Issue #2270, opened 2021-02-19
ibm_is_vpc_routing_table_route does not accept a VPN connection ID as next_hop


hbog:
### Terraform Version
Terraform v0.14.6
+ provider registry.terraform.io/ibm-cloud/ibm v1.21.0

### Affected Resource(s)
- ibm_is_vpc_routing_table_route

### Debug Output
https://gist.github.com/hbog/6a54e8a3a8688c23f3f37d41c6a8e8d1

### Expected Behavior
A route should be created with the vpn gateway connection as next-hop, as per the API documentation:
--next-hop value: If the action is 'deliver', the IP address or VPN connection ID of the next hop to which to route packets

### Actual Behavior
The next-hop only accepts an IP address. A valid VPN connection ID is not accepted and throws an error: Error: Invalid json payload provided 

### Steps to Reproduce

```hcl
resource "ibm_is_vpc" "vpc1" {
  name = "vpc1"
}
resource "ibm_is_subnet" "subnet1" {
  name            = "subnet1"
  vpc             = ibm_is_vpc.vpc1.id
  zone            = var.zone
  ipv4_cidr_block = "10.240.0.0/28"
}
resource "ibm_is_ipsec_policy" "example" {
  name                     = "test-ipsec"
  authentication_algorithm = "md5"
  encryption_algorithm     = "triple_des"
  pfs                      = "disabled"
}
resource "ibm_is_vpn_gateway" "VPNGateway1" {
  name   = "vpn1"
  subnet = ibm_is_subnet.subnet1.id
  mode   = "route"
}
resource "ibm_is_vpn_gateway_connection" "VPNGatewayConnection1" {
  name          = "vpnconn1"
  vpn_gateway   = ibm_is_vpn_gateway.VPNGateway1.id
  peer_address  = ibm_is_vpn_gateway.VPNGateway1.public_ip_address
  preshared_key = "VPNDemoPassword"
  local_cidrs   = [ibm_is_subnet.subnet1.ipv4_cidr_block]
  peer_cidrs    = "10.10.0.0/16"
  ipsec_policy  = ibm_is_ipsec_policy.example.id
}
resource "ibm_is_vpc_routing_table_route" "route-example-dc" {
  routing_table = ibm_is_subnet.subnet1.routing_table
  name        = "route-example-dc"                         
  vpc         = ibm_is_vpc.vpc1.id                   
  zone        = var.zone                        
  destination = "10.10.0.0/16"
[...]

PR #2409, opened 2021-03-25
Add vpn connection as next hop for a vpc routing table route

astha-jain:
Screen Shot 2021-03-24 at 11 58 32 PM

Screen Shot 2021-03-24 at 11 58 59 PM


astha-jain:
Closes #2270

PR #2412, opened 2021-03-25
[Enhancement] lb update supporting sg change

SunithaGudisagarIBM:
@hkantare Can you please review this PR

### DIFF
--- a/ibm/resource_ibm_is_lb.go
+++ b/ibm/resource_ibm_is_lb.go
@@ -112,7 +112,6 @@ func resourceIBMISLB() *schema.Resource {
 			isLBSecurityGroups: {
 				Type:          schema.TypeSet,
 				Computed:      true,
-				ForceNew:      true,
 				Optional:      true,
 				Elem:          &schema.Schema{Type: schema.TypeString},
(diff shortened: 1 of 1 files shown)


ibm/resource_ibm_is_lb.go
```diff
@@ -722,6 +732,40 @@ func lbUpdate(d *schema.ResourceData, meta interface{}, id, name string, hasChan
...
+		if len(remove) > 0 {
+			for _, d := range remove {
+				getSecurityGroupTargetOptions := &vpcv1.GetSecurityGroupTargetOptions{
+					SecurityGroupID: &d,
+					ID:              &id,
+				}
+				_, response, err := sess.GetSecurityGroupTarget(getSecurityGroupTargetOptions)
+				if err != nil {
+					if response != nil && response.StatusCode == 404 {
+						return nil
```
hkantare, line 757: Here should it be continue instead of return nil?
since we should check remaining elements

SunithaGudisagarIBM: @hkantare PR comments oncorporated , Thank you very much. Can you please check now :)

PR #2427, opened 2021-03-29
Datasource support for iam service policy, policies updated

VaishnaviGopal:


hkantare:
You cann't edit the existing resource just by removing the arguments...Need to generate a new datasource

PR #2438, opened 2021-03-30
added tag regex validation for all supported resources

ujjwal-ibm:
added tag regex validation for all other resources supporting tag creation

### DIFF
--- a/ibm/resource_ibm_container_cluster.go
+++ b/ibm/resource_ibm_container_cluster.go
@@ -396,7 +396,7 @@ func resourceIBMContainerCluster() *schema.Resource {
 				Type:        schema.TypeSet,
 				Optional:    true,
 				Computed:    true,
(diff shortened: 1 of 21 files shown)


ibm/resource_ibm_container_cluster.go
```diff
@@ -396,7 +396,7 @@ func resourceIBMContainerCluster() *schema.Resource {
 				Type:        schema.TypeSet,
 				Optional:    true,
 				Computed:    true,
-				Elem:        &schema.Schema{Type: schema.TypeString},
+				Elem:        &schema.Schema{Type: schema.TypeString, ValidateFunc: InvokeValidator("ibm_is_vpc", "tag")},
```
hkantare, line 399: I think here you cann't use ibm_is_vpc in all these we need to add the resourceInvoke validator and add each respective resource name

Issue #1880, opened 2020-09-11
Feature Request: Configure COS bucket retention policy


maristgeek:
Hi there,

We would like to be able to configure a retention policy on a COS bucket through terraform rather than hand manipulation.

https://cloud.ibm.com/docs/cloud-object-storage?topic=cloud-object-storage-immutable

maristgeek:
Semi-related to #1590.

ileneruth:
Is there any update on this one?

kavya498:
Available in [latest](https://github.com/IBM-Cloud/terraform-provider-ibm/releases/tag/v1.23.2)

kavya498:
Closing this issue
Thanks.

PR #2449, opened 2021-03-31
added support for cos retention

Priyaranimishra:
https://github.com/IBM-Cloud/terraform-provider-ibm/issues/1880

### DIFF
--- a/ibm/data_source_ibm_cos_bucket.go
+++ b/ibm/data_source_ibm_cos_bucket.go
@@ -186,6 +186,44 @@ func dataSourceIBMCosBucket() *schema.Resource {
 					},
 				},
 			},
+			"retention_rule": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1000,
+				Description: "A retention policy is enabled at the IBM Cloud Object Storage bucket level. Minimum, maximum and default retention period are defined by this policy and apply to all objects in the bucket.",
+				ForceNew:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"default": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validateAllowedRangeInt(0, 365243),
+							Description:  "If an object is stored in the bucket without specifying a custom retention period.",
+							ForceNew:     false,
+						},
+						"maximum": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validateAllowedRangeInt(0, 365243),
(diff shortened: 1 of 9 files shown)


ibm/data_source_ibm_cos_bucket.go
```diff
@@ -186,6 +186,44 @@ func dataSourceIBMCosBucket() *schema.Resource {
...
 				},
 			},
+			"retention_rule": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1000,
+				Description: "A retention policy is enabled at the IBM Cloud Object Storage bucket level. Minimum, maximum and default retention period are defined by this policy and apply to all objects in the bucket.",
+				ForceNew:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
hkantare, line 196: already the higher level attribute indicates its retention do we still need default_retention (_retention) at the end of each
I think default, minimum maximun permanent shd be good enough

ibm/resource_ibm_cos_bucket.go
```diff
@@ -434,6 +472,99 @@ func resourceIBMCOSUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	//// Update  the Retention policy
```
hkantare, line 475: can we move the logic of s3 client intialization outside if block because it some common in many if blocks

ibm/resource_ibm_cos_bucket.go
```diff
@@ -673,6 +804,17 @@ func resourceIBMCOSRead(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	retentionInput := &s3.GetBucketProtectionConfigurationInput{
+		Bucket: aws.String(bucketName),
+	}
+	retentionptr, err := s3Client.GetBucketProtectionConfiguration(retentionInput)
+	if err != nil {
```
hkantare, line 811: Do we need to add if (err != nil && !strings.Contains(err.Error(), "NoSuchLifecycleConfiguration: The lifecycle configuration does not exist")) && (err != nil && bucketPtr != nil && bucketPtr.Firewall != nil && !strings.Contains(err.Error(), "AccessDenied: Access Denied")) {
similar to expiry and other rule....may be can you try creating a cos bucket along with allowd_ip and retention policy and test it

ibm/data_source_ibm_cos_bucket.go
Priyaranimishra: default in **testAccCheckIBMCosBucket_retention()** function taking as a keyword in **resource_ibm_cos_bucket_test.go** . So I kept it as it is.

Issue #2446, opened 2021-03-31
Dedicated host with status 'failed' throws error during destroy


deepaksibm:
### Terraform Version
0.14.8

### Affected Resource(s)

- ibm_is_dedicated_host

### Terraform Configuration Files
```hcl
data "ibm_resource_group" "default" {
  name = "Default"
}

resource "ibm_is_dedicated_host_group" "dh_group01" {
  family = "balanced"
  class = "bx2d"
  zone = "eu-de-1"
  name = "dp-dh-group-bx2d"
  resource_group = data.ibm_resource_group.default.id
}


data "ibm_is_dedicated_host_group" "dgroup" {
	name = ibm_is_dedicated_host_group.dh_group01.name
}

resource "ibm_is_dedicated_host" "is_dedicated_host" {
  profile = "bx2d-host-152x608"
  name = "dp-dedicated-host-bx2d"
host_group = data.ibm_is_dedicated_host_group.dgroup.id
  
```

### Debug Output
Error: DedicatedHost instance is in FAILED state

### Expected Behavior
Dedicated host should be deleted without throwing error

### Actual Behavior
Deletion fails throwing error saying the State is FAILED

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

kavya498:
Fixed in latest [1.23.0](https://github.com/IBM-Cloud/terraform-provider-ibm/releases/tag/v1.23.0)
thanks..

PR #2452, opened 2021-03-31
issue 2446 dedicated host destroying on failed state handled

deepaksibm:
#2446

### DIFF
--- a/ibm/resource_ibm_is_dedicated_host.go
+++ b/ibm/resource_ibm_is_dedicated_host.go
@@ -510,7 +510,7 @@ func resourceIbmIsDedicatedHostDelete(context context.Context, d *schema.Resourc
 
 	getDedicatedHostOptions.SetID(d.Id())
 
-	_, response, err := vpcClient.GetDedicatedHostWithContext(context, getDedicatedHostOptions)
+	dedicatedHost, response, err := vpcClient.GetDedicatedHostWithContext(context, getDedicatedHostOptions)
 	if err != nil {
 		if response != nil && response.StatusCode == 404 {
 			d.SetId("")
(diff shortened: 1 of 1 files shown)


ibm/resource_ibm_is_dedicated_host.go
```diff
@@ -520,17 +520,18 @@ func resourceIbmIsDedicatedHostDelete(context context.Context, d *schema.Resourc
...
-	updateDedicatedHostOptions := &vpcv1.UpdateDedicatedHostOptions{}
-	dedicatedHostPrototypeMap := map[string]interface{}{}
-	dedicatedHostPrototypeMap["instance_placement_enabled"] = core.BoolPtr(false)
-	updateDedicatedHostOptions.SetID(d.Id())
-	updateDedicatedHostOptions.SetDedicatedHostPatch(dedicatedHostPrototypeMap)
-	_, updateresponse, err := vpcClient.UpdateDedicatedHostWithContext(context, updateDedicatedHostOptions)
-	if err != nil {
-		log.Printf("[DEBUG] UpdateDedicatedHostWithContext failed %s\n%s", err, updateresponse)
-		return diag.FromErr(err)
+	if *dedicatedHost.LifecycleState != isDedicatedHostSuspended && *dedicatedHost.LifecycleState != isDedicatedHostFailed {
```
kavya498, line 523: Lifecycle state cannot be suspended and failed at once..so it should be || not && condition.. and also check for nil pointers
`if dedicatedHost!=nil && dedicatedHost.LifecycleState!=nil&&(*dedicatedHost.LifecycleState != isDedicatedHostSuspended ||*dedicatedHost.LifecycleState != isDedicatedHostFailed ){`

deepaksibm: Will add the nil checks. With || if the state is not suspended but is failed, it will still enter the loop and do update.

PR #2458, opened 2021-04-02
While creating a DL Connect gateway do not wait for gateway to be provisioned  for few providers

ajay-malhotra1:
When a DL Connect gateway is created we wait for the gateway to be moved to provisioned state. But for few providers we shall not wait for the gateway to be provisioned.

So we will check the provider to which the port, that is being used to create a connect gateway, belongs and if it matches those providers we will not wait for the gateway to be provisioned.


ibm/resource_ibm_dl_gateway.go
```diff
@@ -541,10 +542,20 @@ func resourceIBMdlGatewayCreate(d *schema.ResourceData, meta interface{}) error
 
 	log.Printf("[INFO] Created Direct Link Gateway (%s Template) : %s", dtype, *gateway.ID)
 	if dtype == "connect" {
-		_, err = isWaitForDirectLinkAvailable(directLink, d.Id(), d.Timeout(schema.TimeoutCreate))
+		getPortOptions := directLink.NewGetPortOptions(*gateway.Port.ID)
+		port, response, err := directLink.GetPort(getPortOptions)
 		if err != nil {
```
hkantare, line 547: if err != nil  || response != nil || port != nil

so that we are safe to handle the condition if port is nil by any chance

PR #2538, opened 2021-04-23
Limit & KeyID Feature for Keys DataSource

harshit777:
**Current Behavior**
The KP Terraform data source doesn't have provision to search all keys in an instance for a given key-name as an input. Moreover, there is a limitation in KP SDK which limits the number of keys to 2000 when it is not overwritten. Due to this behaviour, there are chances that the user may not be able to retrieve the required key when searching based on the key name and hence this behaviour need to be updated with the following changes so that the user has the flexibility to include all the keys to search for the required key name and also able to search based on key_id & key_alias.

**Expected Behavior**
- [X] Optional param to send the limit and add pagination logic to search until the limit when it is passed in the input. 
- [X] key_id or key_alias as optional input
- [X] Both datasource_ibm_kms_keys and datasource_ibm_kms_key to be updated with these changes
- [X] When the limit is not passed, the behaviour is as it is today.


website/docs/d/kms_key.html.markdown
```diff
@@ -31,20 +37,24 @@ resource "ibm_cos_bucket" "flex-us-south" {
 }
 ```
 
-**NOTE : Data of the key can be retrieved either using a key name or an alias name (if created for the key or keys) .
+**NOTE :
+1) Data of the key can be retrieved either using a key name or an alias name (if created for the key or keys) .
```
hkantare, line 41: (if created for the key or keys) what do you mean by this not clear

website/docs/d/kms_key.html.markdown
```diff
@@ -31,20 +37,24 @@ resource "ibm_cos_bucket" "flex-us-south" {
...
+2) limit is an optional parameter used with the keyname, which iterates and fetches the key till the limit given. When the limit is not passed then the first 2000 keys are fetched according to SDK default behaviour.
```
hkantare, line 42: Is the default value 2000 or 200 in code we see it as 200

website/docs/d/kms_key.html.markdown
```diff
@@ -31,20 +37,24 @@ resource "ibm_cos_bucket" "flex-us-south" {
...
-* `alias` - (Required, In conflict with key_name, string) The alias name associated with the key. Only the key with matching alias name will be retreived.
+* `key_id` - (Required, In conflict with alias_name,key_name, string) The keyID of the key to be fetched.
+* `limit` - (Optional, int) The limit till the keys need to be fetched in the instance.
+* `key_name` - (Required, In conflict with alias_name,key_id string) The name of the key. Only the keys with matching name will be retreived.
+* `alias` - (Required, In conflict with key_name,key_id string) The alias name associated with the key. Only the key with matching alias name will be retreived.
 * `endpoint_type` - (Optional, string) The type of the endpoint (public or private) to be used for fetching keys.
 
 ## Attribute Reference
 
-In addition to all arguments above, the following attributes are exported:
```
hkantare, line 47: Please take the latest docs changes

website/docs/d/kms_key.html.markdown
```diff
@@ -31,20 +37,24 @@ resource "ibm_cos_bucket" "flex-us-south" {
...
 ## Argument Reference
 
 The following arguments are supported:
 
 * `instance_id` - (Required, string) The keyprotect instance guid.
-* `key_name` - (Required, In conflict with alias_name, string) The name of the key. Only the keys with matching name will be retreived.
-* `alias` - (Required, In conflict with key_name, string) The alias name associated with the key. Only the key with matching alias name will be retreived.
+* `key_id` - (Required, In conflict with alias_name,key_name, string) The keyID of the key to be fetched.
+* `limit` - (Optional, int) The limit till the keys need to be fetched in the instance.
+* `key_name` - (Required, In conflict with alias_name,key_id string) The name of the key. Only the keys with matching name will be retreived.
```
hkantare, line 51: please take latest doc and update the argumets and attributes are arranged in alphabetical order

website/docs/d/kms_keys.html.markdown
```diff
@@ -31,9 +31,13 @@ The following arguments are supported:
 
 * `instance_id` - (Required, string) The keyprotect instance guid.
 * `key_name` - (Optional, string) The name of the key. Only the keys with matching name will be retreived.
+* `key_id` - (Required, In conflict with alias_name,key_name, string) The keyID of the key to be fetched.
+* `limit` - (Optional, int) The limit till the keys need to be fetched in the instance.
```
hkantare, line 35: same as above doc

website/docs/d/kms_key.html.markdown
harshit777: Default value is 2000 if the user doesn't specify the limit but  if the limit is specified, the GetKey API fetches the Keys with a range of 200 keys per call

PR #2547, opened 2021-04-27
UI-14859: Instance Group Delete Bugfix

mayukh-sarkar-1:
The bug was while deleting the instance_groups, we need to check if the load balancer is active or not. If not, then wait for the load balancer to be active and then delete the instance groups. Otherwise raise the error. 

Jira Task: https://jiracloud.swg.usma.ibm.com:8443/browse/UI-14859

**Bug Description**
Destroy fails with
 
```
{{Error: Error Deleting the InstanceGroup: Load Balancer r006-c66bd695-f88b-4c7d-a4a5-ed69637d021b provisioning_status must be active to delete instance group
{
"StatusCode": 409,
"Headers":
{ ..., "X-Request-Id": [ "9bc944a6-9297-448c-97e1-742c0d1856da" ], "X-Xss-Protection": [ "1; mode=block" ] }
,
"Result": {
"errors": [
{
"code": "conflict_field",
"message": "Load Balancer r006-c66bd695-f88b-4c7d-a4a5-ed69637d021b provisioning_status must be active to delete instance group",
"target":
{ "name": "load_balancer", "type": "load_balancer" }
}
],
"trace": "9bc944a6-9297-448c-97e1-742c0d1856da"
},
"RawResult": null
}}}
```

### DIFF
--- a/ibm/resource_ibm_is_lb.go
+++ b/ibm/resource_ibm_is_lb.go
@@ -985,7 +985,7 @@ func isWaitForLBAvailable(sess *vpcv1.VpcV1, lbId string, timeout time.Duration)
 	log.Printf("Waiting for load balancer (%s) to be available.", lbId)
 
 	stateConf := &resource.StateChangeConf{
-		Pending:    []string{"retry", isLBProvisioning, "update_pending", "delete_pending"},
+		Pending:    []string{"retry", isLBProvisioning, "update_pending"},
 		Target:     []string{isLBProvisioningDone, ""},
 		Refresh:    isLBRefreshFunc(sess, lbId),
 		Timeout:    timeout,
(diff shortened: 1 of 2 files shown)


ibm/resource_ibm_is_lb.go
```diff
@@ -985,7 +985,7 @@ func isWaitForLBAvailable(sess *vpcv1.VpcV1, lbId string, timeout time.Duration)
 	log.Printf("Waiting for load balancer (%s) to be available.", lbId)
 
 	stateConf := &resource.StateChangeConf{
-		Pending:    []string{"retry", isLBProvisioning},
+		Pending:    []string{"retry", isLBProvisioning, "update_pending", "delete_pending"},
```
deepaksibm, line 988: Update_pending states are not needed , and the refresh function returns "provisioning" for all other states except for 'active' and 'failed'

mayukh-sarkar-1: Hi @deepaksibm I am using the load_balancer's resource function. So I can't remove anything from there because it is used for load_balancer creation as well.

hkantare: why delete_pending is addeed for active status

mayukh-sarkar-1: @hkantare Added just like that..Should I remove it??

PR #2552, opened 2021-04-28
added support to make the prefix as default in vpc

### LINKED ISSUE #2282: No way to make a vpc_address_prefix default on vpc when using manual address_prefix_management
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
v0.13.6

### Affected Resource(s)
- ibm_is_vpc
- ibm_is_vpc_address_prefix

### Terraform Configuration Files
```hcl
resource "ibm_is_vpc" "vpc1" {
  name = "${var.basename}-vpc"
  resource_group = var.resource_group
  address_prefix_management = "manual"
}

resource "ibm_is_vpc_address_prefix" "vpc1_address_prefix_zone1" {
  count = var.address_prefix_management == "manual" ? 1 : 0
  name = "${var.basename}-address-prefix-zone1"
  zone   = local.ZONE1
  vpc    = ibm_is_vpc.vpc1.id
  cidr   = var.vpc_address_range_zone1
}

resource "ibm_is_vpc_address_prefix" "vpc1_address_prefix_zone2" {
  count = var.address_prefix_management == "manual" ? 1 : 0
  name = "${var.basename}-address-prefix-zone2"
  zone   = local.ZONE2
  vpc         = ibm_is_vpc.vpc1.id
  cidr        = var.vpc_address_range_zone2
  
# WORKAROUND FOR ISSUE ----> Set the address prefixes as the default - If we don't do this step that we cannot create
# subnets because there's no default address prefix

# Note the "split" function call - this is because the id returned from creating the address
# comes back as <vpc_id>/<address_range_id> and the update call wants these passed as separate
# arguments.  I suspect this is actually a defect in what is returned from ibm_is_vpc_address_prefix
# and it may one day be fixed and trip up this code.
resource "null_resource" "post_vpc_address_pfx_default" {
  depends_on = [null_resource.post_vpc_rename_defaults]

  count = var.address_prefix_management == "manual" ? 1 : 0

  provisioner "local-exec" {
    command = <<COMMAND
      ibmcloud login --apikey ${var.ibmcloud_api_key} -r ${var.region} -g ${var.resource_group} --quiet ; \
      ibmcloud is vpc-address-prefix-update ${ibm_is_vpc.vpc1.id} ${split("/", ibm_is_vpc_address_prefix.vpc1_address_prefix_zone1[count.index].id)[1]} --default true
      ibmcloud is vpc-address-prefix-update ${ibm_is_vpc.vpc1.id} ${split("/", ibm_is_vpc_address_prefix.vpc1_address_prefix_zone2[count.index].id)[1]}  --default true
     COMMAND
  }
}
```

### Expected Behavior
There should be a way to specify that an address_prefix is default.  Without this, then we end up with no defaults
when we use manual mode.

### Actual Behavior
I end up with no default address_prefixes in the vpc and cannot create subnets.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

ujjwal-ibm:
Relates OR Closes #2282

```
--- PASS: TestAccIBMISVPCAddressPrefix_basic (192.30s)
PASS
ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm	196.098s

--- PASS: TestAccIBMISVPCAddressPrefix_InvalidCidr (0.53s)
PASS
ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm	3.251s

...
```

### DIFF
--- a/ibm/resource_ibm_is_vpc_address_prefix.go
+++ b/ibm/resource_ibm_is_vpc_address_prefix.go
@@ -17,6 +17,7 @@ const (
 	isVPCAddressPrefixCIDR       = "cidr"
 	isVPCAddressPrefixVPCID      = "vpc"
 	isVPCAddressPrefixHasSubnets = "has_subnets"
+	isVPCAddressPrefixDefault    = "is_default"
 )
 
 func resourceIBMISVpcAddressPrefix() *schema.Resource {
@@ -49,6 +50,12 @@ func resourceIBMISVpcAddressPrefix() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: InvokeValidator("ibm_is_address_prefix", isVPCAddressPrefixCIDR),
 				Description:  "CIDIR address prefix",
+			},
+			isVPCAddressPrefixDefault: {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Is default prefix for this zone in this VPC",
 			},
 
 			isVPCAddressPrefixVPCID: {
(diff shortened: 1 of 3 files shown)


ibm/resource_ibm_is_vpc_address_prefix.go
```diff
@@ -303,14 +317,17 @@ func resourceIBMISVpcAddressPrefixUpdate(d *schema.ResourceData, meta interface{
 		name = d.Get(isVPCAddressPrefixPrefixName).(string)
 		hasChanged = true
 	}
-
+	if d.HasChange(isVPCAddressPrefixDefault) {
```
hkantare, line 320: I think the update will override the value of is_default...Let take scenario user created intially with is_default true...after successfull apply he just update the name....
isDefault := false this value will be passed to update even there is no change

ujjwal-ibm: added a separate check for isDefault and name

PR #2570, opened 2021-05-04
enhancement: removed headers and raw result from response for error scenario

ujjwal-ibm:
new error response sample:

```
 Error: Error while creating VPC {
│     "Message": "Provided Name (xyz) is not unique",
│     "StatusCode": 400,
│     "Result": {
│         "errors": [
│             {
│                 "code": "validation_unique_failed",
│                 "message": "Provided Name (xyz) is not unique",
│                 "target": {
│                     "name": "name",
│                     "type": "field",
│                     "value": "xyzzy"
│                 }
│             }
│         ],
│         "trace": "xyz"
│     }
│ }
...
```

### DIFF
--- a/ibm/structures.go
+++ b/ibm/structures.go
@@ -2146,6 +2146,37 @@ func flatterSatelliteZones(zones *schema.Set) []string {
 	return zoneList
 }
 
+// error object
+type ServiceErrorResponse struct {
+	Message    string
+	StatusCode int
+	Result     interface{}
+}
+
+func beautifyError(err error, response *core.DetailedResponse) *ServiceErrorResponse {
+	var (
+		statusCode int
+		result     interface{}
+	)
+	if response != nil {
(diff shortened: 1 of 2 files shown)


ibm/structures.go
```diff
@@ -1961,3 +1962,26 @@ func flatterSatelliteZones(zones *schema.Set) []string {
...
+	StatusCode int
+	Result     interface{}
+}
+
+func beautifyError(err error, response *core.DetailedResponse) *BeautifiedDetailedResponse {
+	return &BeautifiedDetailedResponse{
+		Message:    err.Error(),
+		StatusCode: response.StatusCode,
+		Result:     response.Result,
+	}
```
astha-jain, line 1978: Also capture traceID. Thanks

ujjwal-ibm: trace is in the result

Issue #2307, opened 2021-03-03
docs for ibm_is_virtual_endpoint_gateway are not sufficient


powellquiring:
Here is an example for cos:

```
resource "ibm_is_virtual_endpoint_gateway" "cos" {
  vpc            = module.transits["transit"].vpc.id
  name            = "${module.transits["transit"].vpc.name}-cos"
  resource_group = data.ibm_resource_group.group.id
  target {
    # crn           = ibm_resource_instance.cos.id
    crn           = "crn:v1:bluemix:public:cloud-object-storage:global:::endpoint:s3.direct.dal.us.cloud-object-storage.appdomain.cloud"
    # name = "cloud-object-storage-s3-direct-us-dal-us-south"
    resource_type = "provider_cloud_service"
  }

  # one Reserved IP for per zone in the VPC
  dynamic "ips" {
    for_each = { for subnet in module.transits["transit"].subnets : subnet.id => subnet }
    content {
      subnet = ips.key
      name   = "${ips.value.name}-cos"
    }
  }
}
```

The `ibmcloud is endpoint-gateway-create --help` gave me the clue that I needed to figure this out:

```
$ ibmcloud is endpoint-gateway-create --help
NAME:
    endpoint-gateway-create - Create an endpoint gateway

USAGE:
    ibmcloud is endpoint-gateway-create --vpc-id VPC_ID --target TARGET [--name NAME] [(--reserved-ip-id RESERVED_IP_ID1 --reserved-ip-id RESERVED_IP_ID2 ...) | (--new-reserved-ip NEW_RESERVED_IP1 --new-reserved-ip NEW_RESERVED_IP2 ...)] [--resource-group-id RESOURCE_GROUP_ID | --resource-group-name RESOURCE_GROUP_NAME] [--output JSON] [-q, --quiet]

EXAMPLE:
...
OPTIONS:
    --target value               The name of the provider infrastructure service or the CRN for a provider cloud service instance. You can use command 'ibmcloud is endpoint-gateway-targets' to list the provider cloud and infrastructure services that are qualified to be set as endpoint gateway target.
    --name value                 New name of the endpoint gateway
```

The docs need some additional examples, COS for sure and ntp

```
resource "ibm_is_virtual_endpoint_gateway" "ntp" {
  target {
    name           = "ibm-ntp-server"
    resource_type = "provider_infrastructure_service"
  }
```

argeiger:
Agree with @powellquiring.  

[...]

Issue #2549, opened 2021-04-28
ibm_is_public_gateway selected by subnet


ThomasUllrich666:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
If you want to assign a subnet to an already existing public gateway you need to set the id of the public gateway. You can not easily create a new one because you just can have one by each zone (https://cloud.ibm.com/docs/vpc-on-classic-network?topic=vpc-on-classic-network-about-networking-for-vpc#use-a-public-gateway). This means that you can use a ibm_is_public_gateway data source. Unfortunately you need the name of the public gateway as a required parameter. Due to the fact that this name can set by other terraform scripts or sub projects it is hard to figure this out. 

A better idea would be that you select it by the name of the subnet. (you have a 1:1 relation)

### Potential Terraform Configuration

```hcl
data "ibm_is_public_gateway" "public_gateway" {
  for_each = var.pam_locations
  vpc      = data.ibm_is_vpc.pam_vpc.id
  zone     = each.key

}
```

* #0000

mayukh-sarkar-1:
Hi...Please check if this [PR](https://github.com/IBM-Cloud/terraform-provider-ibm/pull/2571) solves your problem!

astha-jain:
@kavya498 please close this issue. work item complete

PR #2592, opened 2021-05-07
Updated the public gateway docs as per the actual output and code

mayukh-sarkar-1:
Impacts #2549 but may not closes it.


astha-jain:
Closes #2307

PR #2610, opened 2021-05-11
Add new API param to provider

Daniel-Byrne:
Update provider to include a new parameter `max_sessions_per_identity` in the [Account Settings API](https://cloud.ibm.com/apidocs/iam-identity-token-api#updateaccountsettings)

Relevant tests are passing.
Screenshot 2021-05-11 at 15 28 45

### DIFF
--- a/ibm/data_source_ibm_iam_account_settings.go
+++ b/ibm/data_source_ibm_iam_account_settings.go
@@ -6,9 +6,10 @@ package ibm
 import (
 	"context"
 	"fmt"
+	"log"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"log"
 
 	"github.com/IBM/platform-services-go-sdk/iamidentityv1"
 )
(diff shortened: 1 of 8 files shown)


ibm/data_source_ibm_iam_account_settings.go
```diff
@@ -117,7 +123,6 @@ func dataSourceIbmIamAccountSettingsRead(context context.Context, d *schema.Reso
 	}
 
 	getAccountSettingsOptions := &iamidentityv1.GetAccountSettingsOptions{}
-	//getAccountSettingsOptions.SetIncludeHistory(d.Get("include_history").(bool))
```
hkantare, line 120: Why include_history is removed from option struct...we have it has a optioanl argument

website/docs/d/iam_account_settings.html.markdown
```diff
@@ -47,3 +47,4 @@ In addition to all arguments above, the following attributes are exported:
 
 * `session_invalidation_in_seconds` - Defines the period of time in seconds in which a session will be invalidated due  to inactivity. Valid values:   * Any whole number between '900' and '7200'   * NOT_SET - To unset account setting and use service default.
 
+* `max_sessions_per_identity` - Defines the max allowed sessions per identity required by the account. Value values: * Any whole number greater than '0'   * NOT_SET - To unset account setting and use service default
```
hkantare, line 50: This description of what values should go into resource doc...datasource we just set back the value which we get from API

ibm/data_source_ibm_iam_account_settings.go
Daniel-Byrne: This was commented out originally, but I have re-added it now and included it in the list of args in the md file.

website/docs/d/iam_account_settings.html.markdown
Daniel-Byrne: Thanks, I've moved these to the resource doc.

PR #2909, opened 2021-07-27
AppID Token Config Data Source and Resource

dariusbakunas:
## New AppID Token Config Data Source and Resource

AppID token customization: https://cloud.ibm.com/docs/appid?topic=appid-customizing-tokens 

This is initial PR for adding AppID resources. We've worked on unofficial AppID provider that is currently being used by multiple teams in production: https://registry.terraform.io/providers/dariusbakunas/appid/latest/docs

This provider is using official AppID management go sdk client: https://github.com/IBM/appid-management-go-sdk

### Deletion behavior:

When this resource is deleted, it resets AppID token configuration to AppID defaults, which is:

```json
{
    "access": {
        "expires_in": 3600
    },
    "refresh": {
        "enabled": false,
        "expires_in": 2592000
    },
    "anonymousAccess": {
        "enabled": true,
        "expires_in": 2592000
    },
    "accessTokenClaims": [],
    "idTokenClaims": [],
}
```

### Community Note

```
$ make testacc TESTARGS='-run=TestAccIBMAppIDTokenConfigDataSource_basic'

=== RUN   TestAccIBMAppIDTokenConfigDataSource_basic
--- PASS: TestAccIBMAppIDTokenConfigDataSource_basic (55.49s)
PASS

```


vladimiratanasov (approved):

ibm/resource_ibm_appid_token_config.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	tenantID := d.Id()
+
+	tokenConfig, _, err := appidClient.GetTokensConfigWithContext(ctx, &appid.GetTokensConfigOptions{
+		TenantID: &tenantID,
+	})
+
+	if err != nil {
+		return diag.Errorf("Error reading AppID token configuration: %s", err)
```
hkantare, line 153: Can we handle the situation when user deletes this configuration outside the TF using the cli or API...so on
then we may get 404 so if response is not nil and status code is 404 just set d.seID("") and retrun nil
```
if err != nil {
			d.SetId("")
			return nil
		}
		return diag.Errorf("Error reading AppID token configuration: %s", err)
	}
	```

File: website/docs/d/appid_token_config.html.markdown [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+```terraform
+data "ibm_appid_token_config" "tc" {
+    tenant_id = var.tenant_id
+}
+```
+
+## Argument reference
+Review the argument references that you can specify for your data source.
+
+- `tenant_id` - (String) The AppID instance GUID
```
hkantare, line 23: For arguments section can you add its required/optional
 eg tenant_id` - (Requires, String) The AppID instance GUID

dariusbakunas: I was wondering if you considered using https://github.com/hashicorp/terraform-plugin-docs in the future? I know this is alpha at the moment, but I was already getting decent results for unoficial AppID provider :) anyway, I will add these manual changes as requested

ibm/resource_ibm_appid_token_config.go
dariusbakunas: Theoretically it is not possible to delete `token configuration` although it is possible to delete whole AppID instance, in that case you would get 404, which I believe would be handled by another resource that creates AppID instance, ie instance would be re-created and thus tenantID would change. When tenantID changes all of AppID resources would force reload. I can still add this logic just in case.

ibm/config.go
```diff
@@ -1159,6 +1169,24 @@ func (c *Config) ClientSession() (interface{}, error) {
 		}
 	}
 
+	appIDEndpoint := fmt.Sprintf("https://%s.appid.cloud.ibm.com", c.Region)
```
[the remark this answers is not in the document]
dariusbakunas: Let me ask AppID team on this one, honestly I'm not sure, will let you know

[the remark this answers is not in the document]
dariusbakunas: I just got confirmation - AppID does not have private endpoints

PR #2976, opened 2021-08-11
fix: delete topic does not fail when topic does not exist

JunliWang:
the issue is: 
https://ibm-cloudplatform.slack.com/archives/C53NF5MB4/p1628689859157700

the fix includes:
1. increase the default timeout from 3s to 30s
2. when topic does not exist, consider topic deletion is successful
3. bump dependencies(sarama) and broker version from 2.3 to 2.6


hkantare:
Can you please resolve the conflicts in go.sum

PR #3015, opened 2021-08-19
feat: add debugger integration

dariusbakunas:
Terraform supports using debuggers like `delve` to debug providers, this is following their examples on how to integrate it:

https://www.terraform.io/docs/extend/debugging.html

### Community Note

Relates OR Closes #0000

This does not affect any resources or data sources, it only adds `debug` flag support in main function

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -15,6 +15,9 @@ tools:
 build: fmtcheck vet
 	go install
 
+build-dbg:
+	go install -gcflags "all=-N -l"
+
 bin: fmtcheck vet tools
 	@TF_RELEASE=1 sh -c "'$(CURDIR)/scripts/build.sh'"
 
--- a/main.go
+++ b/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"context"
+	"flag"
 	"log"
 
 	"github.com/IBM-Cloud/terraform-provider-ibm/ibm"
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -9,8 +11,24 @@ import (
...
+	flag.BoolVar(&debugMode, "debug", false, "set to true to run the IBM terraform provider with support for debuggers like delve")
+	flag.Parse()
+
+	opts := &plugin.ServeOpts{
 		ProviderFunc: ibm.Provider,
-	})
+	}
+
+	if debugMode {
+		err := plugin.Debug(context.Background(), "IBM-Cloud/ibm", opts)
```
kavya498, line 26: should it not be `registry.terraform.io/IBM-Cloud/ibm` instead of `IBM-Cloud/ibm`?

GNUmakefile
```diff
@@ -15,6 +15,9 @@ tools:
 build: fmtcheck vet
 	go install
 
+build-dbg:
```
kavya498, line 18: why is this required in MAKEfile?

main.go
dariusbakunas: I tested with `IBM-Cloud/ibm` but I can test with full url too

GNUmakefile
dariusbakunas: you need to disable compiler optimizations to produce debuggable binary, so you need to pass thos `gcflags`. It is easier to remember `make build-dbg` vs `go install -gcflags "all=-N -l"`

main.go
dariusbakunas: Looks like it works both ways, so I changed to full url

dariusbakunas:
Maybe I should've opened this as feature request first? :) I don't mind closing it if you don't think this would be useful at this point, but I find it is sometimes easier to look at values during runtime instead of trying to find issue with logs, either way, open for suggestions

PR #3047, opened 2021-09-01
Decoupled Key Policies from Key Resource

harshit777:
Decoupled Kms_Key_Policy from Kms_Key resource


harshit777:
DO NOT MERGE YET !

hkantare:
Can you test one more scenario ...May be provision the resources using old binary (like provision key and some policies) now update the binary to this new one and run terraform plan (it should not show any diff in infrastructure)

hkantare:
closing the PR

Issue #2275, opened 2021-02-23
Provider produced inconsistent result after apply when creating event stream topic 


sflaherty2009:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.13.6

### Affected Resource(s)
Please list the resources as a list, for example:
-ibm_event_streams_topic

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "ibm_event_streams_topic" "clima_matched_conditions" {
  resource_instance_id = ibm_resource_instance.es_instance.id
  name                 = "clima-matched-conditions"
  partitions           = 10
  config = {
    "cleanup.policy"  = "compact,delete"
    "retention.ms"    = "86400000"
    "retention.bytes" = "1073741824"
    "segment.bytes"   = "536870912"
  }
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the 
```
Error: Provider produced inconsistent result after apply

When applying changes to ibm_event_streams_topic.clima_matched_conditions,
provider "registry.terraform.io/ibm-cloud/ibm" produced an unexpected new
value: Root resource was present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

### Expected Behavior
Created a list of topics for the IBM Event Stream

### Actual Behavior
Only created some of the topics. 

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

hkantare:
We are bot able to reproduce it can you let us know the stpes you followed 
```
ibm_resource_instance.es_instance: Creation complete after 21s [id=crn:v1:bluemix:public:messagehub:us-south:a/883079c85357a1f3f85d968780e56518:164bfab0-9284-4ee4-bb5d-cc4c1fba4180::]
ibm_event_streams_topic.es_topic: Creating...
2021/02/24 18:05:58 [DEBUG] ibm_event_streams_topic.es_topic: applying the planned Create change
[...]

PR #3223, opened 2021-10-18
fix: Event Streams: do not return error when topic exists in creation

JunliWang:
when terraform state is out of sync with topic state, error could happen when re-apply the configuration.

https://github.com/IBM-Cloud/terraform-provider-ibm/issues/2275


hkantare:
@JunliWang any update on PR

sflaherty2009:
@JunliWang Do we have any update on this?

sflaherty2009:
@JunliWang @hkantare Do we have any update on this. This is something that breaks at random when building our cluster through terraform and creates a lot of manual work in order to correct.

PR #3319, opened 2021-11-11
adding custom retry to fix the enabling of logging and moniotirg

kavya498:
### Community Note


ocofaigh:
Did anyone actually find out why the error `A logging or monitoring configuration for this cluster already exists. To use a different configuration, delete the existing configuration and try again` was being thrown? How do we know adding a retry is the solution here? 409 is usually a conflict -  it doesn't sound like a retry is going to help here, unless someone knows the root cause to the error ??

PR #3358, opened 2021-11-24
Consistency issue

harshit777:
Updating the Client Config before every resource Provisioning to avoid consistency.

### DIFF
--- a/ibm/config.go
+++ b/ibm/config.go
@@ -701,6 +701,29 @@ func (sess clientSession) keyProtectAPI() (*kp.Client, error) {
 }
 
 func (sess clientSession) keyManagementAPI() (*kp.Client, error) {
+
+	if sess.kmsErr == nil {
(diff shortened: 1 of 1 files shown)


kavya498:
Please reduce number of commits and resolve conflicts

ibm/config.go
```diff
@@ -689,6 +689,20 @@ func (sess clientSession) keyProtectAPI() (*kp.Client, error) {
 }
 
 func (sess clientSession) keyManagementAPI() (*kp.Client, error) {
+
+	if sess.kmsErr == nil {
+		clientConfig := &kp.ClientConfig{
+			BaseURL: envFallBack([]string{"IBMCLOUD_KP_API_ENDPOINT"}, sess.kmsAPI.Config.BaseURL),
```
hkantare, line 695: I think here we need to check two ways...
if APIkey then create session with APIkey else use the bearer token

Issue #3351, opened 2021-11-21
`ibm_kms_key_policies` crashes


elmarx:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* ibm_kms_key_policies

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
resource "ibm_kms_key" "root" {
  instance_id = ibm_resource_instance.kms_instance.id
  key_name    = "root"
}

resource "ibm_kms_key_policies" "root" {
  instance_id = ibm_resource_instance.kms_instance.id
  key_id      = ibm_kms_key.root.key_id
  rotation {
    interval_month = 3
  }
  dual_auth_delete {
    enabled = false
  }
}

```

### Debug Output

https://gist.github.com/elmarx/e5865e4d31e54cfe9531586d8426ec90#file-debug-log

### Panic Output

https://gist.github.com/elmarx/e5865e4d31e54cfe9531586d8426ec90#file-crash

### Expected Behavior

The key policy should be applied as specified, the plugin should not crash.

### Actual Behavior

The plugin crashes. Checking the web-interface the settings did not apply.

### Steps to Reproduce

1. `terraform apply`

gvdibm:
@elmarx bug fix was delivered on 9th Dec, 2021. So, can you verify and close this issue?

PR #3413, opened 2021-12-09
Bug Fix Policy

harshit777:
Bug Fix for Policy crash when utilising CRN for instance_id #3351


ibm/resource_ibm_kms_key_policies.go
```diff
@@ -204,7 +204,7 @@ func resourceIBMKmsKeyPolicyCreate(context context.Context, d *schema.ResourceDa
 		return diag.Errorf("Could not create policies: %s", err)
 	}
 	d.SetId(key.CRN)
-	return resourceIBMKmsKeyPolicyUpdate(context, d, meta)
```
kavya498, line 207: Why is it changed to Read?

hkantare: Actually it was by mistake added to update after create..IT should go to read

PR #3709, opened 2022-04-04
Consolidate power image argument/attribute constants

ChaseRun:
Move image argument/attribute constants defined in the power-go-client into ibm_pi_constants.go.

Files affected:
 - data_source_ibm_pi_catalog_images.go
 - data_source_ibm_pi_image.go
 - data_source_ibm_pi_images.go
 - resource_ibm_pi_image.go
 - resource_ibm_pi_image_export.go

### Community Note

Relates OR Closes #3704

### DIFF
--- a/ibm/service/power/ibm_pi_constants.go
+++ b/ibm/service/power/ibm_pi_constants.go
@@ -61,4 +61,58 @@ const (
 	PIVPNConnectionDeadPeerDetectionThreshold = "threshold"
 	PIVPNConnectionLocalGatewayAddress        = "local_gateway_address"
 	PIVPNConnectionVpnGatewayAddress          = "gateway_address"
+
+	PICloudInstanceID = "pi_cloud_instance_id"
+
+	// ---------- Image ----------
+	// Arguments
+	PIImageName                  = "pi_image_name"
+	PIImageAffinityInstance      = "pi_affinity_instance"
+	PIImageAffinityPolicy        = "pi_affinity_policy"
+	PIImageAffinityVolume        = "pi_affinity_volume"
+	PIImageAntiAffinityInstances = "pi_anti_affinity_instances"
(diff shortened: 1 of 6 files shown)


ibm/service/power/ibm_pi_constants.go
```diff
@@ -0,0 +1,52 @@
...
+	PIImageSecretKey             = "pi_image_secret_key"
+	PIImageStoragePool           = "pi_image_storage_pool"
+	PIImageStorageType           = "pi_image_storage_type"
+	CatalogImagesSAP             = "sap"
+	CatalogImagesVTL             = "vtl"
+
+	// Attributes
+	Images               = "image_info"
+	CatalogImages        = "images"
+	ImagesID             = "id"
```
rmoralesjr, line 29: For common strings like id and name do you want just use the same constants rather than have the same strings defined over and over for the different resource types?

ChaseRun: For now I think it's better to redefine the strings. We can easily review and see all arguments/attributes for a resource type.

ibm/service/power/resource_ibm_pi_image.go
```diff
@@ -339,8 +338,8 @@ func isWaitForIBMPIImageAvailable(ctx context.Context, client *st.IBMPIImageClie
 	log.Printf("Waiting for Power Image (%s) to be available.", id)
 
 	stateConf := &resource.StateChangeConf{
-		Pending:    []string{"retry", helpers.PIImageQueStatus},
-		Target:     []string{helpers.PIImageActiveStatus},
+		Pending:    []string{"retry", "queued"},
+		Target:     []string{"active"},
```
rmoralesjr, line 342: What isn't a constant used here like the previous version?

PR #3715, opened 2022-04-07
Added CRUD operation for sat in cos

Priyaranimishra:
Need to check for Object versioning and existing:--

1. I am able to enable object versioning in tfstate and seeing Status: ENABLED in read operation ....though it is not entering to Flatten method.....and UI still show disabled.

2. If bucket "t1" already exist in cos instance UI and we will try to create from terraform....it did not throw error "Bucket already exists" instead of this able to create the bucket with same name...

```
versioningData := flex.FlattenCosObejctVersioning(versionPtr)
log.Println("Read ..Set versiononh data .....",versioningData)
if len(versioningData) > 0 {
            d.Set("object_versioning", versioningData)
            log.Println("versioning is.....",versioningData)
        } else {
			d.Set("object_versioning", nil)
			log.Println("Read versioning is null..")
		}
}
```

```
Output:--- 
 Read ..Set versiononh data ..... [map[enable:true]]: timestamp=2022-04-06T11:09:27.019Z
versioning is..... [map[enable:true]]: timestamp=2022-04-06T11:09:27.019Z
```

### DIFF
--- a/examples/ibm-cos-bucket/main.tf
+++ b/examples/ibm-cos-bucket/main.tf
@@ -1,114 +1,132 @@
-data "ibm_resource_group" "cos_group" {
-  name = var.resource_group_name
-}
-resource "ibm_resource_instance" "cos_instance" {
-  name              = "cos-instance"
-  resource_group_id = data.ibm_resource_group.cos_group.id
-  service           = "cloud-object-storage"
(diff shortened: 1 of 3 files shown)


hkantare:
closing this PR

PR #3779, opened 2022-05-13
Remove Power Cloud Instance helpers

ChaseRun:
Removed power-go-client constants from Cloud Instance files.
Defined arguments and attributes in ibm_pi_constants.go.

### DIFF
--- a/ibm/service/power/ibm_pi_constants.go
+++ b/ibm/service/power/ibm_pi_constants.go
@@ -4,21 +4,29 @@ import "time"
 
 const (
 
-	// Keys
-	PIKeys    = "keys"
-	PIKeyName = "name"
-	PIKey     = "ssh_key"
-	PIKeyDate = "creation_date"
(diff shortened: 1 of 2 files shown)


ibm/service/power/ibm_pi_constants.go
```diff
@@ -4,21 +4,29 @@ import "time"
...
-	PISAPProfileCertified = "certified"
-	PISAPProfileCores     = "cores"
-	PISAPProfileMemory    = "memory"
-	PISAPProfileID        = "profile_id"
-	PISAPProfileType      = "type"
+	// -- Cloud Connection ----------------------------------------------------
+
+	// -- Cloud Instance ------------------------------------------------------
+	// Required Arguments
+	Arg_CloudInstanceID = "pi_cloud_instance_id"
```
hkantare, line 13: Some are following with Arg_XXX Att_XX for repsective argument/attributes but some are like PISAPProfiles...
can we have some common standard to represent

PR #3852, opened 2022-06-17
Deprecate Whitelist for IBM-cloud-databases. Introduce allowlisting

omaraibrahim:
In the api version 5 whitelist was deprecated and allowlist was introduced. In our strive for feature parity, in the following PR I deprecated whitelist and introduced allowlist.

This involved:
- Deprecating whitelist in our data_source_ibm_database and introducing allowlist
- Deprecating whitelist in our database resource and introducing allowlist
- Updating tests

### Community Note

Relates OR Closes #0000

```
$ make testacc TEST=./ibm/service/database TESTARGS='-run=TestAccIBMDatabaseDataSource_basic'
--- PASS: TestAccIBMDatabaseDataSource_basic (597.39s)
PASS
ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/database	598.703s
...
```

```
$ make testacc TEST=./ibm/service/database TESTARGS='-run=TestAccIBMEDBDatabaseInstanceBasic'
--- PASS: TestAccIBMEDBDatabaseInstanceBasic (1836.72s)
PASS
ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/database	1837.595s
```

```
$ make testacc TEST=./ibm/service/database TESTARGS='-run=TestAccIBMDatabaseInstancePostgresBasic'
--- PASS: TestAccIBMDatabaseInstancePostgresBasic (1493.88s)
PASS
ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/database	1494.758s
```

```
$ make testacc TEST=./ibm/service/database TESTARGS='-run=TestAccIBMDatabaseInstancePostgresAllowlistMigration'
--- PASS: TestAccIBMDatabaseInstancePostgresAllowlistMigration (665.06s)
PASS
ok      github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/database        668.257s
```


ibm/service/database/data_source_ibm_database.go
```diff
@@ -700,11 +720,26 @@ func dataSourceIBMDatabaseInstanceRead(d *schema.ResourceData, meta interface{})
 	}
 	d.Set("auto_scaling", flattenICDAutoScalingGroup(autoSclaingGroup))
 
-	whitelist, err := icdClient.Whitelists().GetWhitelist(icdId)
-	if err != nil {
-		return fmt.Errorf("[ERROR] Error getting database whitelist: %s", err)
+	if _, ok := d.GetOk("whitelist"); ok {
```
hkantare, line 723: whitelist and allowlist are computed variables you can't do d.GetOk...

hkantare: can we use cloud v5 API and flatten the result back to both whitelist and allowlist

ibm/service/database/resource_ibm_database.go
```diff
@@ -459,6 +459,29 @@ func ResourceIBMDatabaseInstance() *schema.Resource {
 						},
 					},
 				},
+				Deprecated:    "Whitelist is deprecated please use allowlist",
+				ConflictsWith: []string{"allowlist"},
+			},
+			"allowlist": {
```
hkantare, line 465: Can't we set the v5 API response of allowlist to whitelist (because the schema looks same for both) it just flatten and setting the same

ibm/service/database/resource_ibm_database.go
```diff
@@ -1710,11 +1763,25 @@ func resourceIBMDatabaseInstanceRead(context context.Context, d *schema.Resource
 	}
 	d.Set("auto_scaling", flattenICDAutoScalingGroup(autoSclaingGroup))
 
-	whitelist, err := icdClient.Whitelists().GetWhitelist(icdId)
-	if err != nil {
-		return diag.FromErr(fmt.Errorf("[ERROR] Error getting database whitelist: %s", err))
+	if _, ok := d.GetOk("whitelist"); ok {
```
hkantare, line 1766: In Resource Read also don't use d.Get operations because the import functionality doesn't work

ibm/service/database/resource_ibm_database.go
omaraibrahim: Hi @hkantare!

They're two different apis endpoints, and despite them being very similar, I think it's safer to keep them separate. Whitelist is depreciated and will be removed soon as well. Hence, I think a separation of concerns is better. Here are the api docs for both. https://github.com/IBM-Cloud/terraform-provider-ibm/pull/3852#discussion_r902894675

Thanks!
Omar

ibm/service/database/resource_ibm_database.go
hkantare: Thanks for explanation
getOK does work in resource read  but it fails during the import scenario. Let say you have created  a database outside Terraform and want to import and  manage in future using Terraform. We need to import the resource. (https://learn.hashicorp.com/tutorials/terraform/state-import?in=terraform/state&utm_source=WEBSITE&utm_medium=WEB_IO&utm_offer=ARTICLE_PAGE&utm_content=DOCS)
For import we first create a empty resource block
resource "ibm_database" "mydb" {
}
Run terraform import ibm_database.mydb <id>
Import also calls the same resource read since the block is empty d.GetOk fails for both whitelist and allowlist

ibm/service/database/resource_ibm_database.go
hkantare: Yes ...v4/v5 API are two different endpoint with same model structure and functionality ...cant we just handle in the backend (i mean resource/datasource) to migrate to V5 AP's without any change to schema.

ibm/service/database/resource_ibm_database.go
omaraibrahim: @hkantare thank you very very much for the explanation :D I've gone ahead and update the  code so that if whitelist isn't there we always use allowlist. I follow the terraform docs here: https://www.terraform.io/plugin/sdkv2/best-practices/deprecations#renaming-an-optional-attribute

So if `d.getOK("whitelist")` fails we'll always use allowlist. lmk what you think :D

ibm/service/database/resource_ibm_database.go
omaraibrahim: @hkantare so the responses are the same yes :D But the supporting functionality is not. Meaning, the autogenerated allowlist functions utilize allowlist structs the use pointers. For example, https://github.com/IBM/cloud-databases-go-sdk/blob/de91b81e136fd4c53c0000e87932aaf150658a33/clouddatabasesv5/cloud_databases_v5.go#L1848

In contrast the whitelist functions utilizes a whitelist struct that does not use pointers. An example of the functions that consume structs that are shaped differently can  be found here: https://github.com/IBM-Cloud/bluemix-go/blob/d538cb4fd9bece50f7b012496df63b3f62d58662/api/icd/icdv4/whitelist.go#L38

So despite achieving the same result, the code is structured very differently and the structs that allowlist relies on are different than whitelist. For example, `AllowlistEntry` is written as follows: https://github.com/IBM/cloud-databases-go-sdk/blob/de91b81e136fd4c53c0000e87932aaf150658a33/clouddatabasesv5/cloud_databases_v5.go#L2046
While `WhitelistEntry` is written as follows https://github.com/IBM-Cloud/bluemix-go/blob/d538cb4fd9bece50f7b012496df63b3f62d58662/api/icd/icdv4/whitelist.go#L13

My concern of just setting the v5 response of allowlist to whitelist is we'll have to do lots of downstream changes that will break things or cause unexpected bugs.

omaraibrahim: Oh I also test the instances using import and it works :)

hkantare:
@omaraibrahim Can you add a migration testcase from whitelist to allowlist

omaraibrahim:
@hkantare I've gone ahead and added migration tests. I've also modified the code to address some concerns and ensure a seemless migration to `allowlist` - while retaining full functionality for `whitelist`.

Issue #3838, opened 2022-06-09
enable coreos support for IBM Cloud Satellite location creation


bhpratt:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

IBM Cloud Satellite now support coreos host OS for control plane and worker nodes. Requires a new location.

### New or Affected Resource(s)

TBD. I think this only affects satellite location create.

* ibm_XXXXX

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* support has been added to the go-sdk: https://github.com/IBM-Cloud/container-services-go-sdk/pull/4#issuecomment-1151096890

bhpratt:
Looks like I can't assign users - but I'd like to take this issue.

darrellschrag:
Also need to make sure you can get the attach CoreOS ignition script from the location once this capability has been added.

andrewferrier:
As @darrellschrag says, I think this affects `ibm_satellite_location` resource and `ibm_satellite_attach_host_script` data source.

andrewferrier:
Looks like the change for `ibm_satellite_location` has gone in, but the one for `ibm_satellite_attach_host_script` is still pending?

andrewferrier:
@hkantare which version of the provider contains this change to the host attach script?

hkantare:
We are planning for a beta release mostly by tomrrow EOD

PR #3914, opened 2022-07-06
Add CoreOS-enabled option for Satellite location create

bhpratt:
### Community Note

Relates #3838

```
Running tool: /usr/local/go/bin/go test -timeout 30s -run ^TestAccSatelliteLocation_Basic$ github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/satellite

ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/satellite	0.892s

```

```
make testacc TEST=./ibm/service/satellite TESTARGS='-run=TestAccSatelliteLocation_Basic'                
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./ibm/service/satellite -v -run=TestAccSatelliteLocation_Basic -timeout 700m

....

[INFO] Set the environment variable IBM_CLUSTER_ID for ibm_container_vpc_worker_pool resource or datasource else tests will fail if this is not set correctly
=== RUN   TestAccSatelliteLocation_Basic
--- PASS: TestAccSatelliteLocation_Basic (1108.65s)
PASS
```

Note, does require updating container services SDK to latest: `github.com/IBM-Cloud/container-services-go-sdk v0.0.0-20220622142911-811d18c8c775` to pick up SDK updates.

Related SDK PRs:

https://github.com/IBM-Cloud/container-services-go-sdk/pull/4#issue-1239122637
https://github.com/IBM-Cloud/container-services-go-sdk/pull/5/files


ibm/service/satellite/resource_ibm_satellite_location.go
```diff
@@ -98,6 +98,11 @@ func ResourceIBMSatelliteLocation() *schema.Resource {
 				Optional:    true,
 				Description: "A description of the new Satellite location",
 			},
+			"coreos_enabled": {
```
hkantare, line 101: if we don't send coreos_enabled by default will it be false? If so can we mark argument as both optinal and computed
Can we set the argument in Read method ...?

hkantare: Can we add this argument to datasource also as computed

bhpratt: Correct. Default is false.

I have made the changes you requested. Thanks for the review.

Out of curiosity, what does `resourceIBMSatelliteLocationRead` do? It pulls in info on the location - but where is that function used in the provider?

hkantare: resourceIBMSatelliteLocationRead is the function which will be used before any action like apply/update.. It first makes a refresh call(resourceIBMSatelliteLocationRead) and checks whether infrastructure already exists by comparing to statefile(if present)
It writes the data to statfile which helps to maintian the current state of templates..
It is used during import scenario also where he wants to bring the infrastructure which was provisioned outside Terraform and monitor using Terraform in future.
Its a key method which helps to find if any drift exists between the exisiting infrastructure and the data stored in statefile

ibm/service/satellite/data_source_ibm_satellite_location.go
```diff
@@ -36,6 +36,12 @@ func DataSourceIBMSatelliteLocation() *schema.Resource {
 				Computed:    true,
 				Description: "A description of the new Satellite location",
 			},
+			"coreos_enabled": {
```
hkantare, line 39: in datasource the field is only computed ...we are just setting the using d.Set(XXX).
so optional needs to removed
```
			"coreos_enabled": {
				Type:        schema.TypeBool,
				Computed:    true,
				Description: "If Red Hat CoreOS features are enabled within the Satellite location",
			},
			```

hkantare: Add the attribute in respective datasource doc also under Attribute Section
website/docs/d/satellite_location.html.markdown file

ibm/service/satellite/resource_ibm_satellite_location.go
bhpratt: Thank you for the detailed explanation. That's quite helpful.

ibm/service/satellite/data_source_ibm_satellite_location.go
bhpratt: Ok, thanks for the clarification. Updated.

PR #3952, opened 2022-08-04
Host script coreos v2

bhpratt:
### Community Note

Closes https://github.com/IBM-Cloud/terraform-provider-ibm/issues/3838

```
=== RUN   TestAccIBMSatelliteAttachHostScriptDataSourceBasic
--- PASS: TestAccIBMSatelliteAttachHostScriptDataSourceBasic (959.11s)
=== RUN   TestAccIBMSatelliteAttachHostScriptDataSourceBasicCoreos
--- PASS: TestAccIBMSatelliteAttachHostScriptDataSourceBasicCoreos (1190.53s)
PASS
ok      github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/satellite       2150.611s
```


bhpratt:
One point worth discussing - as currently written, `custom_script` will not work with coreos. You can't input lines directly into an ignition file. Formatting will be incompatible.

Eventually I do think we want a way to let users add custom stanzas to their coreos ignition script - but for MVP I have bypassed that and put that in a rhel only section.

bhpratt:
I added the note: `RHEL hosts only` to custom_script in the docs. But is there anything programmatically that we can do to prevent it?

ibm/service/satellite/data_source_ibm_satellite_host_script.go
```diff
@@ -126,51 +131,72 @@ func dataSourceIBMSatelliteAttachHostScriptRead(d *schema.ResourceData, meta int
...
+	var host_os string
+	var coreos_enabled bool
+	if _, ok := d.GetOk("coreos_host"); ok {
+		coreos_enabled = d.Get("coreos_host").(bool)
+		if coreos_enabled {
+			host_os = "RHCOS"
+			createRegOptions.OperatingSystem = &host_os
+			scriptPath = filepath.Join(scriptDir, "addHost.ign")
+		}
+	} else {
```
kavya498, line 151: I suspect if it enters this else case at all.. as GetOk on boolean will always have some value as default value for boolean is false..

Can we use `d.GetOkExists("coreos_host")` instead of `d.GetOk("coreos_host")`

kavya498:
@bhpratt , Can you please attach results of terraform plan, apply and show for both scenarios (with and without coreos) either here or slack me..
Don't want to take risk as it will break existing infra/statefile if something goes wrong in this datasource..

ibm/service/satellite/data_source_ibm_satellite_host_script.go
bhpratt: I tested it both with `coreos_host` = false and no `coreos_host` flag - and both times it provided the correct RHEL script.

If I'm understanding the test: `func testAccCheckIBMSatelliteAttachHostScriptDataSourceConfig` - that should show it, as `coreos_host` is not presented, and it returns the correct script.

The test itself doesn't show the contests of the attach script - but locally I can see the correct scripts provided. I'll upload the results here.

bhpratt:
all right - not sure what happened here when I attempted to squash this to 1 commit. I'll open a new PR.

PR #3958, opened 2022-08-09
Fix duplicate VPC entries in cloud conn create

rmoralesjr:
Fix duplicate VPC entries in cloud connection create
operation.

### Community Note

Relates OR Closes #0000

=== RUN   TestAccIBMPICloudConnectionVPC
--- PASS: TestAccIBMPICloudConnectionVPC (376.75s)
PASS
...
```

### DIFF
--- a/ibm/service/power/resource_ibm_pi_cloud_connection.go
+++ b/ibm/service/power/resource_ibm_pi_cloud_connection.go
@@ -145,6 +145,11 @@ func ResourceIBMPICloudConnection() *schema.Resource {
 				Computed:    true,
 				Description: "GRE auto-assigned source IP address",
 			},
+			PICloudConnectionConnectionMode: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Type of service the gateway is attached to",
+			},
 		},
 	}
 }
@@ -200,8 +205,9 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
 			vpcIds := flex.ExpandStringList(v.(*schema.Set).List())
 			vpcs := make([]*models.CloudConnectionVPC, len(vpcIds))
 			for i, vpcId := range vpcIds {
+				vpcIdCopy := vpcId[0:]
 				vpcs[i] = &models.CloudConnectionVPC{
-					VpcID: &vpcId,
+					VpcID: &vpcIdCopy,
 				}
 			}
 			vpc.Vpcs = vpcs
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: ibm/service/power/resource_ibm_pi_cloud_connection.go [hunk 1]
```diff
@@ -145,6 +145,11 @@ func ResourceIBMPICloudConnection() *schema.Resource {
 				Computed:    true,
 				Description: "GRE auto-assigned source IP address",
 			},
+			PICloudConnectionConnectionMode: {
```
yussufsh, line 148: Could you please add this to `Attribute reference` in the resource doc?

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -200,8 +205,9 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
 			vpcIds := flex.ExpandStringList(v.(*schema.Set).List())
 			vpcs := make([]*models.CloudConnectionVPC, len(vpcIds))
 			for i, vpcId := range vpcIds {
+				vpcIdCopy := vpcId[0:]
```
yussufsh, line 208:
```suggestion
				vpcId := vpcId
```
This single line change should work here.

yussufsh (changes requested):

ibm/service/power/resource_ibm_pi_cloud_connection.go
rmoralesjr: I think setting the variable to itself is confusing.

ibm/service/power/resource_ibm_pi_cloud_connection.go
rmoralesjr: Will do. Thanks

ibm/service/power/resource_ibm_pi_cloud_connection.go
yussufsh: I see it is recommended easier approach for such cases. Ref: http://golang.org/doc/faq#closures_and_goroutines

rmoralesjr: I don't read that as recommended but rather easier than their other example of approaching this scenario which was creating an anonymous function that would use a copy of the string so it will not be overridden. So can I keep this version since even that webpage says it seems odd? Thanks.

yussufsh: I don't see any issue with this approach.

yussufsh (approved):
lgtm

PR #3963, opened 2022-08-11
CIS - Delete filter on deletion of Firewall Rules 

arpit-srivastava-ibm:
Problem statement - 
- On deletion of Firewall Rules, the filter is not removed automatically.

Issue - https://github.ibm.com/NetworkTribe/CIS_Support/issues/2515

### DIFF
--- a/ibm/service/cis/resource_ibm_cis_domain.go
+++ b/ibm/service/cis/resource_ibm_cis_domain.go
@@ -71,12 +71,10 @@ func ResourceIBMCISDomain() *schema.Resource {
 			cisDomainVerificationKey: {
 				Type:     schema.TypeString,
 				Computed: true,
-				Optional: true,
 			},
 			cisDomainCnameSuffix: {
 				Type:     schema.TypeString,
 				Computed: true,
-				Optional: true,
 			},
 		},
 		Create:   resourceCISdomainCreate,
--- a/ibm/service/cis/resource_ibm_cis_domain.go
+++ b/ibm/service/cis/resource_ibm_cis_domain.go
@@ -71,10 +71,12 @@ func ResourceIBMCISDomain() *schema.Resource {
 			cisDomainVerificationKey: {
 				Type:     schema.TypeString,
 				Computed: true,
+				Optional: true,
 			},
 			cisDomainCnameSuffix: {
 				Type:     schema.TypeString,
 				Computed: true,
+				Optional: true,
 			},
 		},
 		Create:   resourceCISdomainCreate,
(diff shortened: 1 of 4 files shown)


ibm/service/cis/resource_ibm_cis_firewall_rules.go
```diff
@@ -238,6 +238,21 @@ func ResourceIBMCISFirewallrulesDelete(context context.Context, d *schema.Resour
 		return diag.FromErr(fmt.Errorf("[ERROR] Error deleting the  custom resolver %s:%s", err, response))
 	}
 
+	if id, ok := d.GetOk(cisFilterID); ok {
```
hkantare, line 241: Can we use same filterID forother rules. If we delete the filterID will not be other firewall rules not effected?
If user create a filter using Terraform is the destroy not handled to remove filter

arpit-srivastava-ibm: No we can not use same filter for other rules. It throws an error on creating a new rule with an already mapped filterID.

arpit-srivastava-ibm: Yes, destroy handles filter deletion but it does not cover old filters.

arpit-srivastava-ibm:
Testcase results - 

```
arpit-mac@Arpits-MacBook-Pro terraform-provider-ibm % make testacc TEST=./ibm/service/cis TESTARGS='-run=TestAccIBMCisFirewallrules_Basic'
=== RUN   TestAccIBMCisFirewallrules_Basic
--- PASS: TestAccIBMCisFirewallrules_Basic (92.94s)
PASS
ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/cis	94.358s
```

ibm/service/cis/resource_ibm_cis_domain.go
```diff
@@ -71,10 +71,12 @@ func ResourceIBMCISDomain() *schema.Resource {
 			cisDomainVerificationKey: {
 				Type:     schema.TypeString,
 				Computed: true,
+				Optional: true,
```
hkantare, line 74: why is this marked as optional we are just setting the value in Read method

ibm/service/cis/resource_ibm_cis_domain.go
```diff
@@ -71,10 +71,12 @@ func ResourceIBMCISDomain() *schema.Resource {
...
 			},
 			cisDomainCnameSuffix: {
 				Type:     schema.TypeString,
 				Computed: true,
+				Optional: true,
```
hkantare, line 79: same as above

PR #4001, opened 2022-09-01
fix: schematics regional endpoints issue

kavya498:
### Community Note

Relates OR Closes https://github.ibm.com/blueprint/product/issues/7432

### DIFF
--- a/ibm/conns/config.go
+++ b/ibm/conns/config.go
@@ -1664,15 +1664,10 @@ func (c *Config) ClientSession() (interface{}, error) {
 	}
 
 	// SCHEMATICS Service
-	schematicsEndpoint := "https://schematics.cloud.ibm.com"
+	// schematicsEndpoint := "https://schematics.cloud.ibm.com"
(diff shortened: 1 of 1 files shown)


JonWoodlief:
Will this allow me to view eu-de workspaces from us-south, and vice versa?

JonWoodlief:
Can you make it configurable as an input on the workspace and workspace outputs datasource?

kavya498:
@JonWoodlief ,
It would not be the right way to do it, as schematics is a regional endpoint..
Client gets initialised only once at provider level..

If you want to do so, you ll have to use provider aliases..

JonWoodlief:
Since we can't predict which region the user will install their workspace into, this means that every single instance of a workspace datasource needs an accompanying provider definition with an alias.

With this solution, if I have a template with 2 workspace data sources, I will have authenticate 3 providers. 1 for the code, and 2 for each of the datasources. That's not an acceptable solution. We are building open source solutions and looking to promote them- these kinds of hacks will reflect poorly on IBM.

PR #4117, opened 2022-10-18
Fix attach script loop for satellite host attach

bhpratt:
Custom code is overwriting part of the attach script in some locations, causing the attach script to fail.

This update provides a change to insert the custom code rather than overwriting a line in the attach script. In addition, it changes else/if loop to case for easier code readability and maintenance.

### Community Note

```
 make testacc TEST=./ibm/service/satellite TESTARGS='-run=TestAccIBMSatelliteAttachHostScriptDataSourceBasic'
=== RUN   TestAccIBMSatelliteAttachHostScriptDataSourceBasic
--- PASS: TestAccIBMSatelliteAttachHostScriptDataSourceBasic (886.52s)
=== RUN   TestAccIBMSatelliteAttachHostScriptDataSourceBasicCoreos
--- PASS: TestAccIBMSatelliteAttachHostScriptDataSourceBasicCoreos (1172.36s)
PASS
ok  	github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/satellite	2060.180s
```

### DIFF
--- a/ibm/service/satellite/data_source_ibm_satellite_host_script.go
+++ b/ibm/service/satellite/data_source_ibm_satellite_host_script.go
@@ -160,10 +160,11 @@ func dataSourceIBMSatelliteAttachHostScriptRead(d *schema.ResourceData, meta int
 		return fmt.Errorf("[ERROR] Error Generating Satellite Registration Script: %s\n%s", err, resp)
 	}
 
-	lines := strings.Split(string(resp), "\n")
+	scriptContent := string(resp)
 
 	//if this is a RHEL host, find insert point for custom code
 	if !coreos_enabled {
+		lines := strings.Split(scriptContent, "\n")
 		var index int
 		for i, line := range lines {
 			if strings.Contains(line, `export OPERATING_SYSTEM`) {
@@ -217,11 +218,9 @@ yum install container-selinux -y
 			}
 		}
 
-		lines = append(lines[:index+1], lines[index:]...)
-		lines[index] = insertionText
+		scriptContent = strings.Join(lines[:index+1], "\n") + insertionText + strings.Join(lines[index+1:], "\n")
 	}
 
-	scriptContent := strings.Join(lines, "\n")
 	err = ioutil.WriteFile(scriptPath, []byte(scriptContent), 0644)
 	if err != nil {
 		return fmt.Errorf("[ERROR] Error Creating Satellite Attach Host Script: %s", err)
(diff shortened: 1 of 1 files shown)


ibm/service/satellite/data_source_ibm_satellite_host_script.go
```diff
@@ -188,30 +194,34 @@ elif [[ "${OPERATING_SYSTEM}" == "RHEL8" ]]; then
...
 `
-					}
-				}
+		default:
+			if script, ok := d.GetOk("custom_script"); ok {
+				insertionText = script.(string)
 			}
 		}
+
+		lines[index] = lines[index] + "\n" + insertionText
```
crhuff-ibm, line 221: I think here we might actually want insertionText -> lines[index] instead of the reverse. Currently we look for `operating_system` export, and want our code to execute immediately after.  If there was 
```
export OPERATING_SYSTEM <- insert after here
if A
then B
```

we would end. up with 

```
export OPERATING_SYSTEM
if A <- index = line for operating system + 1
insertion text
then B
```

crhuff-ibm (changes requested):

ibm/service/satellite/data_source_ibm_satellite_host_script.go
bhpratt: In my test it inserted as expected:

```
export OPERATING_SYSTEM
subscription-manager refresh
# subscription-manager repos --enable rhel-server-rhscl-7-rpms
# subscription-manager repos --enable rhel-7-server-optional-rpms
# subscription-manager repos --enable rhel-7-server-rh-common-rpms
# subscription-manager repos --enable rhel-7-server-supplementary-rpms
# subscription-manager repos --enable rhel-7-server-extras-rpms
set -x
subscription-manager repos --enable rhel-8-for-x86_64-appstream-rpms
subscription-manager repos --enable rhel-8-for-x86_64-baseos-rpms

<snip>

systemctl enable $SERVICENAME


if [[ "${OPERATING_SYSTEM}" != "RHEL7" && "${OPERATING_SYSTEM}" != "RHEL8" ]]; then
	echo "This script is only intended to run with a RHEL7 or RHEL8 operating system. Current operating system ${OPERATING_SYSTEM}."
	exit 1
fi
```


which aligns with when we break - we break and index the line that `export Operating_System` is on:

```
		var index int
 		for i, line := range lines {
 			if strings.Contains(line, `export OPERATING_SYSTEM`) {
 				index = i
 				break
 			}
 ```

bhpratt: (that is a change from the original code...there had been an index + 1)

Issue #4216, opened 2022-12-02
ibm_cm_catalog doens't allow to specify destination resource group id


vbontempi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* ibm_cm_catalog

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl

resource "ibm_cm_catalog" "service_vpe_catalog" {
  label             = "test catalog"
  short_description = "catalog"
  resource_group_id = "<<< resource group id >>>"
  catalog_icon_url = "icon url"
  kind = "offering"
  tags = ["catalog", "tags"]
}

provider "ibm" {
  ibmcloud_api_key = var.ibmcloud_api_key
  region           = "eu-gb"
}

variable "ibmcloud_api_key" {
  description = "APIkey that's associated with the account to use, set via environment variable TF_VAR_ibmcloud_api_key"
  type        = string
  sensitive   = true
}
terraform {
  required_version = ">= 1.0.0"
  required_providers {
    ibm = {
      source  = "ibm-cloud/ibm"
      version = "= 1.48.0"
    }
  }
}
```

### Expected Behavior

Create the catalog instance in the resource_group_id RG

### Actual Behavior

Fails with the following error:

```
Can't configure a value for "resource_group_id": its value will be decided automatically based on the result of applying this configuration.
```

### Steps to Reproduce

1. `terraform plan` or `terraform apply`

### References

doc link showing the field as optional
https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/cm_catalog#argument-reference

vbontempi:
By looking at the source code
https://github.com/IBM-Cloud/terraform-provider-ibm/blob/ad385845a8162f1ead86e8374a33676616272b0d/ibm/service/catalogmanagement/resource_ibm_cm_catalog.go#L100

it seems that the field is missing of `Optional: true` field. It has been removed also from other fields that are expected to be optional

ocofaigh:
[...]

PR #4224, opened 2022-12-07
fix: make resource_group_id optional for catalogs

benbuchanan:
This PR is fixing a regression in the catalog resource.  The `resource_group_id` field should be optional, this was changed when code was regenerated recently and we need to fix it as soon as possible because it is breaking teams. Thanks.

issue here: https://github.com/IBM-Cloud/terraform-provider-ibm/issues/4216


hkantare:
yes we will include in next release

PR #4254, opened 2022-12-20
fix: pass ibm cloud session token along to iaas

dnwe:
Currently, in order to use the IBM provider to manager classic infrastructure resources you must either configure the provider manually with `iam_token` and `iam_refresh_token` variables, or you must provide api-specific credentials via the `iaas_classic_username` and `iaas_classic_api_key` variables. However, it should be reasonable when using `ibmcloud_api_key` for the provider to passthrough the access and refresh tokens received as part of the IBM Cloud session.

It seemed like the original code had been in-place to do that, but it was incorrectly guarded by a non-nil check of
`sess.SoftLayerSession.IAMToken` rather than a check to confirm that explicit iaas credentials hadn't been provided already.

### Community Note

Relates #3405

```

### DIFF
--- a/ibm/conns/config.go
+++ b/ibm/conns/config.go
@@ -1324,7 +1324,8 @@ func (c *Config) ClientSession() (interface{}, error) {
 	}
 	session.bmxUserDetails = userConfig
 
-	if sess.SoftLayerSession != nil && sess.SoftLayerSession.IAMToken != "" {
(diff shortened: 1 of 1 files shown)


dnwe:
@hkantare is this something you could take a look at? I think you worked on the original IAM authentication flows

hkantare:
@dnwe We will review and make it avilable as part of next release (Jan end)

PR #4259, opened 2022-12-22
Fix ibm pi instance update wait for status

### LINKED ISSUE #4258: PowerVS VSI update in-place with 'ibm_pi_instance' times out when VSI status is SHUTDOWN
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform CLI and Terraform IBM Provider Version

```sh
$ terraform -v
Terraform v1.3.6
on linux_amd64
+ provider registry.terraform.io/ibm-cloud/ibm v1.48.0
```

### Affected Resource(s)

* ibm_pi_instance

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
terraform {
  required_providers {
    ibm = {
      source  = "IBM-Cloud/ibm"
      version = "1.48.0"
    }
  }
}

provider "ibm" {
  ibmcloud_api_key = var.ibmcloud_api_key // export IC_API_KEY = "<api key>"
  region           = var.region           // export IBMCLOUD_REGION = "<region>"
  zone             = var.zone             // export IBMCLOUD_ZONE = "<zone>"
}

resource "ibm_pi_instance" "resize-test" {
  pi_cloud_instance_id = var.cloud_instance_id
  pi_instance_name     = var.instance_name
  pi_memory            = var.memory
  pi_processors        = var.processors
  pi_sys_type          = var.sys_type
  pi_proc_type         = var.proc_type
  pi_image_id          = var.image_id
  pi_network { network_id = var.network_id }
}
```

### Debug Output

Note: The gist include output from both the working scenario (VSI in 'ACTIVE'
state) and hang scenario (VSI in 'SHUTOFF' state).

https://gist.github.com/jaywcarman/2dde2d72483861250b668aac7e9d78e6

### Panic Output

No panic occurs.

### Expected Behavior

The ibm_pi_instance update in-place action should complete once the VM status
is no longer 'RESIZE', whether the new state is 'ACTIVE', 'SHUTOFF', or
possibly others (I'm not familiar with all possible states, anything other than
error type states should be OK).

### Actual Behavior

The ibm_pi_instance update in-place action times out polling for the VSI state
to be 'ACTIVE'.

### Steps to Reproduce

The Terraform code required to recreate is available here:
https://github.com/jaywcarman/terraform-provider-ibm/tree/ibm-power-resize/examples/ibm-power-resize

1. Create a PowerVS VSI. See project above. I set variables in a `var.tfvars`
   file, export `IBMCLOUD_API_KEY` as an env var and apply via:

    ```
    terraform apply -var-file var.tfvars -var ibmcloud_api_key="$IBMCLOUD_API_KEY"
    ```

2. Shutdown the PowerVS VSI using `ibmcloud` cli or https://cloud.ibm.com web GUI.

3. Modify the `ibm_pi_instance`'s `pi_memory` and/or `pi_processors`
   attributes. We want to update the VSI in place.

4. Wait an hour for the operation to time out ☹️

### Important Factoids

The resize action is successful in all cases, even when Terroform exits with
`Error: context deadline exceeded`.

### References

* #0000

jaywcarman:
### Community Note

Fixes #4258

### DIFF
--- a/ibm/service/power/resource_ibm_pi_instance.go
+++ b/ibm/service/power/resource_ibm_pi_instance.go
@@ -962,7 +962,7 @@ func performChangeAndReboot(ctx context.Context, client *st.IBMPIInstanceClient,
 }
 
 func isWaitforPIInstanceUpdate(ctx context.Context, client *st.IBMPIInstanceClient, id string) (interface{}, error) {
-	log.Printf("Waiting for PIInstance (%s) to be SHUTOFF AFTER THE RESIZE Due to DLPAR Operation ", id)
+	log.Printf("Waiting for PIInstance (%s) to be ACTIVE or SHUTOFF AFTER THE RESIZE Due to DLPAR Operation ", id)
 
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{"RESIZE", "VERIFY_RESIZE"},
--- a/ibm/service/power/resource_ibm_pi_instance.go
+++ b/ibm/service/power/resource_ibm_pi_instance.go
@@ -641,7 +641,7 @@ func resourceIBMPIInstanceUpdate(ctx context.Context, d *schema.ResourceData, me
 			if err != nil {
 				return diag.Errorf("failed to update the lpar with the change %v", err)
 			}
-			_, err = isWaitForPIInstanceAvailable(ctx, client, instanceID, "OK")
+			_, err = isWaitforPIInstanceUpdate(ctx, client, instanceID)
 			if err != nil {
 				return diag.FromErr(err)
 			}


jaywcarman:
@hkantare

The `TestAccIBMPIInstancesDataSource_basic` and `TestAccIBMPIInstancesDataSource_basic` acceptance tests passed for me when I ran them locally with this patch.

Setup:
```
export IC_API_KEY="$(ic_apikeys jwcarman)"
export IAAS_CLASSIC_API_KEY="$(ic_apikeys jwcarman)"
export IAAS_CLASSIC_USERNAME="jwcarman"
export IC_REGION="mon"
export IC_ZONE="mon01"
export PI_CLOUDINSTANCE_ID="8fa27c40-827c-4568-8813-79b398e9cd27"
export PI_NETWORK_NAME="terraform-test-power"
export PI_IMAGE="8de97d73-95c0-46e9-a509-0b10be36eded"
```

---

`export TESTARGS="-run TestAccIBMPIInstancesDataSource_basic"`:
```
=== RUN   TestAccIBMPIInstancesDataSource_basic
--- PASS: TestAccIBMPIInstancesDataSource_basic (16.21s)
PASS
ok      github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/power   16.298s
```

---

`export TESTARGS="-run TestAccIBMPIInstanceBasic"`:
```
=== RUN   TestAccIBMPIInstanceBasic
--- PASS: TestAccIBMPIInstanceBasic (387.52s)
PASS
ok      github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/power   387.616s
```

rmoralesjr:
@yussufsh The existing code has been like this for 3 years. I wonder if the back end services changed behavior to cause the VM to go to VERIFY_RESIZE and then SHUTDOWN when resizing down? The service broker code does not look changed recently so it may be further down the stack. I'll run some tests with the CLI to see if the root cause is on the server side & not in Terraform and post results here.

rmoralesjr:
OK I see in the logs posted in 4258 that the VM was SHUTDOWN prior to the update operation and so after the update the VM went back to SHUTDOWN state after the resizing states. I reproduced this behavior using the CLI. @yussufsh @dhirendersingh19 I think one option is to check the VM state prior to the update operation and look for that state after the update operation use that as a target state.

jaywcarman:
> Or Jay's change to use the isWaitforPIInstanceUpdate function I think would work as well except it doesn't cover error states like the original call to isWaitForPIInstanceAvailable function.

Since the Pending state is limited to `[]string{"RESIZE", "VERIFY_RESIZE"}` ([code ref](https://github.com/IBM-Cloud/terraform-provider-ibm/blob/master/ibm/service/power/resource_ibm_pi_instance.go#L968)) I think an error would be returned if the lpar returns any error state.

See https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/resource/state.go#L44-L45:
> // If the Refresh function returns a state other than the Target state or one
> // listed in Pending, return immediately with an error.

I don't understand why error states are included as `isWaitForPIInstanceAvailable` Target states. Why wouldn't we want an error returned on errors states?

rmoralesjr:
I think the original coder listed the ERROR state as a target since that is a state from which terraform does not have to wait for completion anymore and they did not want to stop the rest of the terraform config from processing. But if we are covered for the ERROR state by the state.go processing then I am OK with the current change. Yussuf and Dhirender do you have concerns?

rmoralesjr:
@jaywcarman I am going to revert this change since it is causing regressions in the scenarios where the PVM instance is active state. I should have asked you why you shutdown your PVM instance to do the mem/proc change outside terraform (in CLI or GUI)? The PVM instance only needs to be shutdown in whwn mem or proc is greater than maxMemLpar and maxCPULpar . . .

` 
 		if mem > maxMemLpar || procs > maxCPULpar {

			err = performChangeAndReboot(ctx, client, instanceID, cloudInstanceID, mem, procs)
			if err != nil {
				return diag.FromErr(err)
			}`

rmoralesjr:
The problem is the isWaitforPIInstanceUpdate doesn't actually detect the target state of Active since the refresh function never passes that state (only SHUTOFF, RESIZE and "") even though the "ACTIVE" string is listed as a Target state in list . . .
```
func isPIInstanceShutAfterResourceChange(client *st.IBMPIInstanceClient, id string) resource.StateRefreshFunc {
	return func() (interface{}, string, error) {

		pvm, err := client.Get(id)
		if err != nil {
			return nil, "", err
		}

		if *pvm.Status == "SHUTOFF" && pvm.Health.Status == helpers.PIInstanceHealthOk {
			log.Printf("The lpar is now off after the resource change...")
			return pvm, "SHUTOFF", nil
		}

		return pvm, "RESIZE", nil
	}
}
```
In the case where mem and proc is changed and no shutdown is needed, the update operation will timeout since shutdown state is never seen. I don't think the manual step (via cli or GUI )of shutting down the VM is needed since the terraform code in the comment above will perform the reboot if needed.

PR #4338, opened 2023-02-07
wip

attilatabori:
### Community Note

Relates OR Closes #0000

```

...
```

### DIFF
--- a/ibm/service/kubernetes/resource_ibm_container_cluster.go
+++ b/ibm/service/kubernetes/resource_ibm_container_cluster.go
@@ -251,7 +251,7 @@ func ResourceIBMContainerCluster() *schema.Resource {
 
 			"machine_type": {
 				Type:        schema.TypeString,
-				ForceNew:    true,
+				ForceNew:    false,
 				Optional:    true,
 				Description: "Machine type",
 			},
@@ -314,7 +314,7 @@ func ResourceIBMContainerCluster() *schema.Resource {
 			"operating_system": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				ForceNew:    true,
+				ForceNew:    false,
 				Computed:    true,
 				Description: "The operating system of the workers in the default worker pool.",
 			},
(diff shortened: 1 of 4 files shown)


ibm/service/kubernetes/resource_ibm_container_vpc_worker_pool.go
```diff
@@ -232,6 +240,10 @@ func ResourceIBMContainerVPCWorkerPoolValidator() *validate.ResourceValidator {
 
 func resourceIBMContainerVpcWorkerPoolCreate(d *schema.ResourceData, meta interface{}) error {
 
+	if b, ok := d.GetOk("default_worker_pool"); ok && b.(bool) {
+		return resourceIBMContainerVpcWorkerPoolRead(d, meta)
```
z0za, line 244: I'm afraid it won't work as is...
Read uses the id, to get the clusterid and workerpoolid, and we don't have that at this point of time.
fortunately we can get the workerpool by name, so we will need to refactor the read a bit.

ibm/service/kubernetes/resource_ibm_container_vpc_worker_pool.go
```diff
@@ -49,6 +49,14 @@ func ResourceIBMContainerVpcWorkerPool() *schema.Resource {
 					"cluster"),
 			},
 
+			"default_worker_pool": {
+				Type:        schema.TypeBool,
+				Required:    false,
+				ForceNew:    false,
+				Optional:    true,
+				Description: "If true, instead of creating one, the create function will read the the clsuter's \"default\" workerpool ",
+			},
```
z0za, line 58: maybe it can be general. what do you think?
```suggestion
			"import_worker_pool": {
				Type:        schema.TypeBool,
				Required:    false,
				ForceNew:    false,
				Optional:    true,
				Description: "If true, instead of creating one, the create function will try to import an existing workerpool",
			},
```

ibm/service/kubernetes/resource_ibm_container_cluster.go
```diff
@@ -1066,38 +1066,38 @@ func resourceIBMContainerClusterUpdate(d *schema.ResourceData, meta interface{})
...
-
-			_, err = WaitForWorkerAvailable(d, meta, targetEnv)
-			if err != nil {
-				return fmt.Errorf("[ERROR] Error waiting for workers of cluster (%s) to become ready: %s", d.Id(), err)
-			}
-		} else {
-			return fmt.Errorf("[ERROR] The default worker pool does not exist. Use ibm_container_worker_pool and ibm_container_worker_pool_zone attachment resources to make changes to your cluster, such as adding zones, adding worker nodes, or updating worker nodes")
-		}
-	}
+	// if d.HasChange("default_pool_size") && !d.IsNewResource() {
```
lewiseevans, line 1069: Commenting out this block of code will prevent customers from managing the number of workers in their default worker pool, right?

Do we want to stage the delivery of this? First release, just update the `ForceNew`. Second release, remove management of the default worker pool from the cluster resource. Will this allow for a smoother transition for customers?

PR #4348, opened 2023-02-14
fix: add fields to version metadata for onboarding fix

benbuchanan:
Some products will be onboarded with a modules section that was not previously supported by the terraform provider, adding the parameters to the version resource metadata fixes the issue.

There were more metadata fields added by the API than expected, these have all been added to the resources and data sources to support cm onboarding.

Also added `x_auth_token` to the docs for the cm version resource, and created the `cm_validation` docs.

### DIFF
--- a/ibm/service/catalogmanagement/data_source_ibm_cm_offering.go
+++ b/ibm/service/catalogmanagement/data_source_ibm_cm_offering.go
@@ -562,6 +562,107 @@ func DataSourceIBMCmOffering() *schema.Resource {
 													Computed:    true,
 													Description: "Version source URL.",
 												},
+												"working_directory": &schema.Schema{
+													Type:        schema.TypeString,
+													Optional:    true,
+													Computed:    true,
+													Description: "Working directory of source files.",
+												},
+												"example_name": &schema.Schema{
+													Type:        schema.TypeString,
+													Optional:    true,
+													Computed:    true,
+													Description: "Working directory of source files.",
+												},
+												"end_deploy_time": &schema.Schema{
+													Type:        schema.TypeString,
+													Optional:    true,
+													Computed:    true,
+													Description: "The time validation ended.",
+												},
+												"usage": &schema.Schema{
+													Type:        schema.TypeString,
+													Optional:    true,
+													Computed:    true,
+													Description: "Usage text for the version.",
+												},
+												"usage_template": &schema.Schema{
+													Type:        schema.TypeString,
+													Optional:    true,
+													Computed:    true,
+													Description: "Usage text for the version.",
+												},
+												"modules": &schema.Schema{
+													Type:        schema.TypeList,
+													Optional:    true,
+													Computed:    true,
+													Description: "Terraform modules.",
+													Elem: &schema.Resource{
+														Schema: map[string]*schema.Schema{
+															"name": &schema.Schema{
+																Type:        schema.TypeString,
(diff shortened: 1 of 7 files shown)


ibm/service/catalogmanagement/data_source_ibm_cm_offering.go
```diff
@@ -562,6 +562,107 @@ func DataSourceIBMCmOffering() *schema.Resource {
 													Computed:    true,
 													Description: "Version source URL.",
 												},
+												"working_directory": &schema.Schema{
+													Type:        schema.TypeString,
```
hkantare, line 566: In datasource these attributes should be only computed.
We decalred in the schema but are we flattening and setting back?

ibm/service/catalogmanagement/resource_ibm_cm_version.go
```diff
@@ -248,6 +248,107 @@ func ResourceIBMCmVersion() *schema.Resource {
 							Optional:    true,
 							Description: "Version source URL.",
 						},
+						"working_directory": &schema.Schema{
```
hkantare, line 251: Same as above we declared in schema are we using these attributes in create/update?

website/docs/r/cm_version.html.markdown
```diff
@@ -110,6 +110,7 @@ Nested scheme for **metadata**:
 * `target_kinds` - (Optional, Forces new resource, List) Deployment target of the content being onboarded. Current valid values are iks, roks, vcenter, power-iaas, terraform, and vpc-x86. Required for virtual server image for VPC.
 * `target_version` - (Optional, Forces new resource, String) The semver value for this new version, if not found in the zip url package content.
 * `working_directory` - (Optional, Forces new resource, String) Optional - The sub-folder within the specified tgz file that contains the software being onboarded.
+* `x_auth_token` - (Optional, Forces new resource, String) The token to access the tgz in the repo.
```
hkantare, line 113: All those attributes should be added to both docs?

benbuchanan: added to docs

ibm/service/catalogmanagement/data_source_ibm_cm_offering.go
benbuchanan: Changed them to computed only, thanks.

I'm not sure what you mean flattening and setting back? These have to be in the metadata because our API sets certain things in the metadata that aren't specified on a create or update.

ibm/service/catalogmanagement/resource_ibm_cm_version.go
benbuchanan: Changed these to all be computed, so they arent used in create/update

ibm/service/catalogmanagement/data_source_ibm_cm_offering.go
benbuchanan: I changed all of the metadata to be computed, so I think that resolves your question.

ibm/service/catalogmanagement/resource_ibm_cm_version.go
hkantare: User will not able to set the metadata during onboarding process?

benbuchanan:
@hkantare please review again, this fix is urgent for a large customer and I would like to have it merged asap, thanks.

PR #4766, opened 2023-08-25
Retry cloud connection create/update when vpc is unavailable

rachari:
### Community Note

Relates OR Closes #0000

```

...
```

### DIFF
--- a/ibm/service/power/resource_ibm_pi_cloud_connection.go
+++ b/ibm/service/power/resource_ibm_pi_cloud_connection.go
@@ -240,7 +240,7 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
 			err = retryCloudConnectionsVPC(func() (err error) {
 				cloudConnection, cloudConnectionJob, err = client.Create(body)
 				return
-			}, vpcRetryCount, vpcRetryDuration, "create", err)
+			}, "create", err)
 		}
 		if err != nil {
 			log.Printf("[DEBUG] create cloud connection failed %v", err)
@@ -355,7 +355,7 @@ func resourceIBMPICloudConnectionUpdate(ctx context.Context, d *schema.ResourceD
 				err = retryCloudConnectionsVPC(func() (err error) {
 					_, cloudConnectionJob, err = client.Update(cloudConnectionID, body)
 					return
-				}, vpcRetryCount, vpcRetryDuration, "update", err)
+				}, "update", err)
 			}
 			if err != nil {
 				log.Printf("[DEBUG] update cloud connection failed %v", err)
@@ -525,10 +525,10 @@ func resourceIBMPICloudConnectionDelete(ctx context.Context, d *schema.ResourceD
 	return nil
 }
 
-func retryCloudConnectionsVPC(ccVPCRetry func() error, retryCount int, waitDuration time.Duration, operation string, errMsg error) (err error) {
-	for count := 0; count < retryCount; count++ {
+func retryCloudConnectionsVPC(ccVPCRetry func() error, operation string, errMsg error) (err error) {
+	for count := 0; count < vpcRetryCount && errMsg != nil; count++ {
 		log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", errMsg)
-		time.Sleep(waitDuration)
+		time.Sleep(vpcRetryDuration)
 		log.Printf("[DEBUG] retrying cloud connection %s, retry #%v", operation, count+1)
 		err = ccVPCRetry()
 	}
--- a/ibm/service/power/resource_ibm_pi_cloud_connection.go
+++ b/ibm/service/power/resource_ibm_pi_cloud_connection.go
@@ -240,10 +240,12 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
 			err = retryCloudConnectionsVPC(func() (err error) {
 				cloudConnection, cloudConnectionJob, err = client.Create(body)
 				return
-			}, vpcRetryCount, vpcRetryDuration, "create")
+			}, vpcRetryCount, vpcRetryDuration, "create", err)
 		}
-		log.Printf("[DEBUG] create cloud connection failed %v", err)
-		return diag.FromErr(err)
+		if err != nil {
+			log.Printf("[DEBUG] create cloud connection failed %v", err)
+			return diag.FromErr(err)
+		}
 	}
 
 	if cloudConnection != nil {
@@ -353,10 +355,12 @@ func resourceIBMPICloudConnectionUpdate(ctx context.Context, d *schema.ResourceD
 				err = retryCloudConnectionsVPC(func() (err error) {
 					_, cloudConnectionJob, err = client.Update(cloudConnectionID, body)
 					return
-				}, vpcRetryCount, vpcRetryDuration, "update")
+				}, vpcRetryCount, vpcRetryDuration, "update", err)
 			}
-			log.Printf("[DEBUG] update cloud connection failed %v", err)
-			return diag.FromErr(err)
+			if err != nil {
+				log.Printf("[DEBUG] update cloud connection failed %v", err)
+				return diag.FromErr(err)
+			}
 		}
 		if cloudConnectionJob != nil {
 			_, err = waitForIBMPIJobCompleted(ctx, jobClient, *cloudConnectionJob.ID, d.Timeout(schema.TimeoutCreate))
@@ -521,15 +525,12 @@ func resourceIBMPICloudConnectionDelete(ctx context.Context, d *schema.ResourceD
 	return nil
 }
 
-func retryCloudConnectionsVPC(ccVPCRetry func() error, retryCount int, waitDuration time.Duration, operation string) (err error) {
+func retryCloudConnectionsVPC(ccVPCRetry func() error, retryCount int, waitDuration time.Duration, operation string, errMsg error) (err error) {
 	for count := 0; count < retryCount; count++ {
-		log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", err)
-		time.Sleep(time.Minute)
+		log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", errMsg)
+		time.Sleep(waitDuration)
 		log.Printf("[DEBUG] retrying cloud connection %s, retry #%v", operation, count+1)
 		err = ccVPCRetry()
-	}
-	if err != nil {
-		return
 	}
 	return
 }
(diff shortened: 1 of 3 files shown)


rachari:
@rmoralesjr @yussufsh Can you review this?

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -22,6 +23,10 @@ import (
 	"github.com/IBM-Cloud/terraform-provider-ibm/ibm/validate"
 )
 
+var (
+	vpcUnavailable = regexp.MustCompile("503|service unavailable|server is down for maintenance or overloaded")
```
yussufsh, line 27: I have seen 503 for many other APIs (especially instance listing).
Why only this particular? If in future we are planning to handle this for all other resources then I think we should do it in power-go-client instead of here.
cc @rmoralesjr

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -224,10 +229,24 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
 	}
 
 	client := st.NewIBMPICloudConnectionClient(ctx, sess, cloudInstanceID)
+	retryCount := 2
 	cloudConnection, cloudConnectionJob, err := client.Create(body)
 	if err != nil {
-		log.Printf("[DEBUG] create cloud connection failed %v", err)
-		return diag.FromErr(err)
+		for count := retryCount; count >= 0; count-- {
```
yussufsh, line 235: How many times do you want to retry?
If it is 3 as expected then use `count > 0` here. Setting `retryCount := 2` and running loop 3 times does not make sense.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -224,10 +229,24 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
 	}
 
 	client := st.NewIBMPICloudConnectionClient(ctx, sess, cloudInstanceID)
+	retryCount := 2
 	cloudConnection, cloudConnectionJob, err := client.Create(body)
```
yussufsh, line 233: Better write a function and call Create() N no of times. Current logic tries 4 times overall.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -224,10 +229,24 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
...
 	if err != nil {
-		log.Printf("[DEBUG] create cloud connection failed %v", err)
-		return diag.FromErr(err)
+		for count := retryCount; count >= 0; count-- {
+			if err == nil {
```
yussufsh, line 236: What is the use of this check? We are checking err is not nil before loop and also inside loop it is checking and returning.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -333,7 +353,21 @@ func resourceIBMPICloudConnectionUpdate(ctx context.Context, d *schema.ResourceD
...
+					log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", err)
+					time.Sleep(time.Minute)
+					log.Printf("[DEBUG] retrying cloud connection update, retry #%v", (retryCount-count)+1)
+					_, cloudConnectionJob, err = client.Update(cloudConnectionID, body)
+				}
+			}
+			if err != nil {
+				log.Printf("[DEBUG] update cloud connection failed %v", err)
+				return diag.FromErr(err)
+			}
```
yussufsh, line 370: Try to reuse code.

yussufsh (changes requested):

rmoralesjr (commented):
Address Yussuf's comments and I am good.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -497,3 +520,16 @@ func resourceIBMPICloudConnectionDelete(ctx context.Context, d *schema.ResourceD
 	d.SetId("")
 	return nil
 }
+
+func retryCloudConnectionsVPC(ccVPCRetry func() error, retryCount int, waitDuration time.Duration, operation string) (err error) {
```
yussufsh, line 524: What is the use of `waitDuration`? It is sleeping for a minute every time.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -497,3 +520,16 @@ func resourceIBMPICloudConnectionDelete(ctx context.Context, d *schema.ResourceD
...
+	for count := 0; count < retryCount; count++ {
+		log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", err)
```
yussufsh, line 526: `err` is nil in 1st iteration.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -497,3 +520,16 @@ func resourceIBMPICloudConnectionDelete(ctx context.Context, d *schema.ResourceD
...
+		time.Sleep(time.Minute)
+		log.Printf("[DEBUG] retrying cloud connection %s, retry #%v", operation, count+1)
+		err = ccVPCRetry()
+	}
+	if err != nil {
+		return
+	}
```
yussufsh, line 533: There is no need to check if not nil here. Just return err.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -333,6 +349,13 @@ func resourceIBMPICloudConnectionUpdate(ctx context.Context, d *schema.ResourceD
 
 		_, cloudConnectionJob, err := client.Update(cloudConnectionID, body)
 		if err != nil {
+			if vpcUnavailable.Match([]byte(err.Error())) {
+				err = retryCloudConnectionsVPC(func() (err error) {
+					_, cloudConnectionJob, err = client.Update(cloudConnectionID, body)
+					return
+				}, vpcRetryCount, vpcRetryDuration, "update")
+			}
+			log.Printf("[DEBUG] update cloud connection failed %v", err)
```
yussufsh, line 358: Why failure? What if retry operation pass?

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -226,6 +236,12 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
...
 	cloudConnection, cloudConnectionJob, err := client.Create(body)
 	if err != nil {
+		if vpcUnavailable.Match([]byte(err.Error())) {
+			err = retryCloudConnectionsVPC(func() (err error) {
+				cloudConnection, cloudConnectionJob, err = client.Create(body)
+				return
+			}, vpcRetryCount, vpcRetryDuration, "create")
+		}
 		log.Printf("[DEBUG] create cloud connection failed %v", err)
 		return diag.FromErr(err)
```
yussufsh, line 246: Should not return err if one of the retry passes.

yussufsh (changes requested):

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -497,3 +524,13 @@ func resourceIBMPICloudConnectionDelete(ctx context.Context, d *schema.ResourceD
 	d.SetId("")
 	return nil
 }
+
+func retryCloudConnectionsVPC(ccVPCRetry func() error, retryCount int, waitDuration time.Duration, operation string, errMsg error) (err error) {
+	for count := 0; count < retryCount; count++ {
+		log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", errMsg)
+		time.Sleep(waitDuration)
+		log.Printf("[DEBUG] retrying cloud connection %s, retry #%v", operation, count+1)
+		err = ccVPCRetry()
```
yussufsh, line 533: This loop will keep on re-trying the func `retryCount` times. We should break if we get one successful call.

ibm/service/power/resource_ibm_pi_cloud_connection.go
```diff
@@ -226,8 +236,16 @@ func resourceIBMPICloudConnectionCreate(ctx context.Context, d *schema.ResourceD
 	client := st.NewIBMPICloudConnectionClient(ctx, sess, cloudInstanceID)
 	cloudConnection, cloudConnectionJob, err := client.Create(body)
 	if err != nil {
-		log.Printf("[DEBUG] create cloud connection failed %v", err)
-		return diag.FromErr(err)
+		if vpcUnavailable.Match([]byte(err.Error())) {
+			err = retryCloudConnectionsVPC(func() (err error) {
+				cloudConnection, cloudConnectionJob, err = client.Create(body)
+				return
+			}, vpcRetryCount, vpcRetryDuration, "create", err)
```
yussufsh, line 243: Any reason to pass `vpcRetryCount`, `vpcRetryDuration`? They are global const; there is no need to pass if they are the same.

yussufsh (changes requested):

ibm/service/power/resource_ibm_pi_cloud_connection.go
yussufsh: You are using `errMsg != nil` in loop condition but it is never changed in the loop :)

yussufsh: Try below code and test out with fake responses. I prefer that 2nd one.
```
func retryCloudConnectionsVPC(ccVPCRetry func() error, operation string, errMsg error) (err error) {
	for count := 0; count < vpcRetryCount && errMsg != nil; count++ {
		log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", errMsg)
		time.Sleep(vpcRetryDuration)
		log.Printf("[DEBUG] retrying cloud connection %s, retry #%v", operation, count+1)
		errMsg = ccVPCRetry()
	}
    err = errMsg
	return
}
```
OR
```
func retryCloudConnectionsVPC(ccVPCRetry func() error, operation string, errMsg error) error {
	for count := 0; count < vpcRetryCount && errMsg != nil; count++ {
		log.Printf("[DEBUG] unable to get vpc details for cloud connection: %v", errMsg)
		time.Sleep(vpcRetryDuration)
		log.Printf("[DEBUG] retrying cloud connection %s, retry #%v", operation, count+1)
		errMsg = ccVPCRetry()
	}
    return errMsg
}
```

yussufsh: You can also pass errMsg pointer and not have any error return

yussufsh (changes requested):

ibm/service/power/resource_ibm_pi_cloud_connection.go
yussufsh: I just need whatever code is present should work fine and not affect the overall time. Also it should be well tested. If we add low quality code then people tend to copy paste same thing again and again.
We should do it right from the start. Cannot compensate for cloud performance but terraform should not put additional strain.

yussufsh (approved):
lgtm

PR #4782, opened 2023-09-04
Updated schematics sdk to latest v0.2.2

VaishnaviGopal:


ibm/service/schematics/resource_ibm_schematics_action.go
```diff
@@ -1072,14 +1071,6 @@ func resourceIBMSchematicsActionCreate(context context.Context, d *schema.Resour
 		}
 		createActionOptions.SetSettings(settings)
 	}
-	if _, ok := d.GetOk("state"); ok {
```
kavya498, line 1075: Is it similar change like sys_lock? then do we also have to have schema changes for state?

Doc changes for these two attributes are missing.

kavya498 (changes requested):

ibm/service/schematics/resource_ibm_schematics_action.go
kavya498: I see that state is already computed. So ignore. :) 
Make sure the docs are updated right.

PR #4807, opened 2023-09-18
fix(share-iops): Share Iops range fix for dp2

deepaksibm:
removing iops validation as only dp2 profile is supported going forward


astha-jain (approved):
Looks good.

hkantare (approved):

PR #4952, opened 2023-11-27
issue-13603-fix attachment terraform

archanaponnada:
### Community Note

Relates OR Closes #0000

**Issue :** 
In ibm_scc_profile_attachment, notification is made mandatory which is incorrect

**Fix :** 
In ibm_scc_profile_attachment, notification is made optional.
Screenshot 2023-11-27 at 2 19 09 PM

Screenshot 2023-11-27 at 2 19 24 PM

**How will it work :** 
1. Now notification is optional, ie when notification is not given also system will accept.
2. If notification object is given and enabled=false, it will ignore controls object - we need to fix at code level.Tracked under issue#13603
3. If notification object is given  and controls object id given - it will be ignore - we need to fix at code level.Tracked under issue#13603
Screenshot 2023-11-27 at 2 30 43 PM

...
```

### DIFF
--- a/ibm/service/scc/resource_ibm_scc_profile_attachment.go
+++ b/ibm/service/scc/resource_ibm_scc_profile_attachment.go
@@ -109,6 +109,7 @@ func ResourceIbmSccProfileAttachment() *schema.Resource {
 			},
 			"notifications": {
 				Type:        schema.TypeList,
+				MaxItems:    1,
 				Optional:    true,
 				Description: "The request payload of the attachment notifications.",
 				Elem: &schema.Resource{

### CODE REVIEW

File: ibm/service/scc/resource_ibm_scc_profile_attachment.go [hunk 1]
```diff
@@ -109,7 +109,7 @@ func ResourceIbmSccProfileAttachment() *schema.Resource {
 			},
 			"notifications": {
 				Type:        schema.TypeList,
-				Required:    true,
+				Optional:    true,
```
hkantare, line 112: Can we add MaxItems to 1 because I see we were accessing only .0 position element

archanaponnada: Added the maxItems property.
Testcase :
image

PR #4993, opened 2023-12-15
Removing Support For Power VPN Create

michaelkad:
### Community Note

Relates OR Closes #0000

```

...
```


yussufsh:
Wondering if we can deprecate the resource itself if the plan is to remove support for update/delete in future. How about including data sources as well?

michaelkad:
> Wondering if we can deprecate the resource itself if the plan is to remove support for update/delete in future. How about including data sources as well?

@yussufsh We will eventually remove support for update and delete but we still support update and delete for now. Whenever that support of it completely remove on power, we will have to remove the  data sources .

ismirlia (approved):

PR #5049, opened 2024-01-16
Remove old deprecated features from terraform

ismirlia:
This pull request removes all deprecated powervs features that are older than 1 year old, and fixes the existing documentation around that. The only remaining deprecated feature is the subnet jumbo field, since that was deprecated this previous release.


michaelkad (approved):
Lgtm

ismirlia:
Resolved merge conflicts. Please let me know if there is anything else to do before merging @yussufsh and @hkantare .

PR #5069, opened 2024-01-25
fix(routing-table-route) - fix routing table route advertise patch and action 

deepaksibm:


deepaksibm:
@hkantare @kavya498 Please consider this PR for the 1.62 release, thanks

PR #5090, opened 2024-01-31
Fix Satellite Locations recreation when updating, remove forcenew, add computed and applyonce

Blintmester:
Fix test steps:

- I created a satellite location with provider `v1.61.0`
- after creation, checked the default values of `pod_subnet` and `service_subnet`
- updated provider version to `v1.62.0`
- tried to `terraform plan` -> it tried to recreate the whole location
- built the local version of the provider, with custom version tag
- updated provider version to custom version tag
- tried to `terraform plan` -> it said no changes in the resource
- tried to update some fields, and then `terraform plan` -> only wanted to update the fields, not recreate the whole location

### Community Note

Relates OR Closes #0000

```

...
```


libesz:
@bhpratt yep, that was the idea. IMHO changing immutable parameters in general should mean force redeploy, but in case of such a huge resource as a satellite location, it is unlikely that one would really want redeploy everything with a terraform apply. The same was suggested for the IP autodetect method -> it is only effective on create, and once a change is requested on a subsequent apply, the user is notified that the request is ignored.

z0za (approved):

Issue #5168, opened 2024-03-04
PowerVS: ibm_pi_instance: pi_replicants parameter , on destroy fails to clean up., on reapply config changes


surajsbharadwaj:
There is a major bug with **ibm_pi_instance**. resource,
When using **pi_replicants** as parameter:
For example if i pass in value 3 
On destroy it doesnt destroy the second and third instance. Only the first instance is destroyed.

surajsbharadwaj:
On reapply again , to make sure there is no infrastructure changes,
it tries to **rename** the instance and also removes the **placement_group_id** which it created automatically when using **pi_replicants**
```
//  # module.powervs_instance_primary.module.pi_instance.ibm_pi_instance.instance will be updated in-place
  ~ resource "ibm_pi_instance" "instance" {
        id                             = "848b2d38-6264-460b-a09d-7095fb3b3f2e/870053af-38a1-4acf-baa4-507a028a31ec"
      ~ pi_instance_name               = "replica-2-1" -> "replica-2"
      - pi_placement_group_id          = "fb9b38bc-0bbd-4ace-b540-c077399fd39d" -> null
        # (32 unchanged attributes hidden)

        # (3 unchanged blocks hidden)
    }

Plan: 1 to add, 1 to change, 0 to destroy.
```

hkantare:
Shared the issue with power team
[@ismirlia](https://ibm-cloudplatform.slack.com/team/W8FNY9DMZ)

ismirlia:
@surajsbharadwaj can you please share the configuration you are doing?

ismirlia:
I opened: https://github.com/IBM-Cloud/terraform-provider-ibm/pull/5176 to fix the issue.

surajsbharadwaj:
Does it also handle the reapply part ?

ismirlia:
@surajsbharadwaj Let me double on that.

surajsbharadwaj:
Can you support the reapply without any modifications? This is important. This will be included in our automation and published to catalog tile

surajsbharadwaj:
Else change the way data is inserted in the state file. Make a dictionary with the instance name as key and values as list of instances..

ismirlia:
> Else change the way data is inserted in the state file. Make a dictionary with the instance name as key and values as list of instances..

@surajsbharadwaj you are gonna have to give me an example of how that would work in the terraform code because I don't know how to make that work perfectly.

[...]

PR #5176, opened 2024-03-06
Fix deleting instances with replicant attribute set

ismirlia:
When the pi_replicant field is set in terraform it creates a bunch of vms as expected, but terraform destroy does not delete these.

This fixes: https://github.com/IBM-Cloud/terraform-provider-ibm/issues/5168

```
--- PASS: TestAccIBMPIInstanceBasic (680.22s)
PASS

--- PASS: TestAccIBMPIInstanceNetwork (1123.39s)
PASS

--- PASS: TestAccIBMPIInstanceReplicant (1914.63s)
PASS
```


michaelkad:
@ismirlia could you add the acceptance test for this?

yussufsh:
> @ismirlia could you add the acceptance test for this?

Even I would like to see the interesting results :) This has been the way since how many years I don't know.
@ismirlia Also you will need to handle `update` flow for every replicant. Good work btw!

ismirlia:
@yussufsh @michaelkad so right now, update only updates the original instance and not the replicants. Should the goal be to update the replicants as well when any update is made? So this could work for some types of updates potentially, but it couldn't work for others. Like for example you couldn't update the name to match for all of them.

I'm still working on running the tests, got delayed by the reapply quesion @surajsbharadwaj had.

ismirlia:
@yussufsh updating all of them in one go is also not currently how the API/CLI work in that regard either.

ismirlia:
Are there any more comments on this? @yussufsh @hkantare ? Updating every replicant at once is not supported, and is not supported by the API. This PR updates just deletion, when including a replicant, and fixes up some of the instance terraform acceptance tests.

michaelkad (approved):

ibm/service/power/resource_ibm_pi_instance.go
```diff
@@ -840,20 +847,23 @@ func resourceIBMPIInstanceDelete(ctx context.Context, d *schema.ResourceData, me
...
+		client := st.NewIBMPIInstanceClient(ctx, sess, cloudInstanceID)
+		err = client.Delete(instanceID)
+		if err != nil {
+			return diag.FromErr(err)
+		}
 
-	_, err = isWaitForPIInstanceDeleted(ctx, client, instanceID)
-	if err != nil {
-		return diag.FromErr(err)
+		_, err = isWaitForPIInstanceDeleted(ctx, client, instanceID)
```
yussufsh, line 863: If instead of delete-wait and loop again, can we have two loops? One loop will delete all the instances and the other loop will check the deletion status. This will save lots of time.

ismirlia: Made changes.

yussufsh (approved):
lgtm

yussufsh:
> lgtm

@ismirlia if an update is not supported can we add a statement in the doc? I am not sure why updating the replicants is not supported. Should be using the same logic as create/delete.

ismirlia:
> > lgtm
> 
> @ismirlia if an update is not supported can we add a statement in the doc? I am not sure why updating the replicants is not supported. Should be using the same logic as create/delete.

Updating the replicants is not really supported by the semantics of the API. This multi-delete approach is already stretching the supported capabilities of what this resource should do. 

The markdown file needs to be updated? pi_instance.html.markdown?

yussufsh:
I understand but then it is always a possibility that user updates the instance details and expect every replicant is getting updated. Imagine if user changes the memory to reduce cost over 50 instances for example, getting to know after some days that changes did not happen and incurred avoidable expenses.
Anyways I'm also fine adding a warning in instance markdown against replicants, so user understands the risks before using the variable.

yussufsh (approved):
lgtm

PR #5218, opened 2024-03-20
Avoid setting service_endpoints default from the plugin

obai-1:
This avoids setting a default value for `service_endpoints` in the Terraform provider plugin. This allows the default to be set by the backend according to the compliance standard in use at the specific region. 

Users still have the option to specify one of the following values  `public`, `private`, or `public-and-private` for `service_endpoints`. If user leaves `service_endpoints` empty, the default value will be set based on the compliance standard in the region where the instance is being created. Generally, if the region is enabled with FS Cloud/ENS High compliance, then the default would be private. Otherwise, the default would be public. During any update, if you leave `service_endpoints` empty, it will maintain the previously selected value.

### Community Note


omaraibrahim (approved):
LGTM 🥳

hkantare (approved):

PR #5220, opened 2024-03-20
feat: use new standardized structures for errors

dpopp07:
This adds support for the new error formats designed for IBM Cloud services and tools. The support is only added in select VPC resources/ data sources for now. The support comes from wrapping the provider to catch errors and re-format them with the new system, so that the change to the actual resource/data source code has as small of a surface area as possible.

The primary change to resource code comes in the from of updating calls to `fmt.Errorf` to be calls to a new wrapper function, `flex.FmtErrorf`. It accepts the same arguments but instead of wrapping errors coming from the underlying Go SDK so that we retain only the message, it detects those errors and returns them so that we preserve all of the information needed for the new error formats.

With this change, error messages will now include more information about the problem and will include an identifying hash that allows the user to lookup more information about the specific error scenario that occurred.

### Community Note

Relates OR Closes #0000

```

...
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/IBM/event-notifications-go-admin-sdk v0.4.0
 	github.com/IBM/eventstreams-go-sdk v1.4.0
 	github.com/IBM/go-sdk-core/v3 v3.2.4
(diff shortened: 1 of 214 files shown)


dpopp07:
Closing in favor of #5278

PR #5278, opened 2024-04-16
feat: wrap provider to create new error structures

dpopp07:
This commit adds code that wraps the provider in order to catch errors and re-format them with the new system, to minimize the surface area of the changes required for non-generated resource/data source code.

The primary change to resource code comes in the from of updating calls to `fmt.Errorf` to be calls to a new wrapper function, `flex.FmtErrorf`. It accepts the same arguments but instead of wrapping errors coming from the underlying Go SDK so that we retain only the message, it creates instances of the new TerraformProblem type to be propagated to the wrapped provider code. This way, we preserve all of the information needed for the new error formats, including the information from previous errors.

Note: This PR just creates the baseline support. Resource code that returns an `error` (rather than a `diag.Diagnostics` will start to get support for new errors right away, but they will not have all information needed from previous errors in the SDK. For that, individual service teams will need to update their code to use `flex.FmtErrorf` instead of `fmt.Errorf`.

@hkantare 

### Community Note

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/IBM/event-notifications-go-admin-sdk v0.4.0
 	github.com/IBM/eventstreams-go-sdk v1.4.0
 	github.com/IBM/go-sdk-core/v3 v3.2.4
-	github.com/IBM/go-sdk-core/v5 v5.16.5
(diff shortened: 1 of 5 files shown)


hkantare:
Can you fix conflicts in go.mod and go.sum files

PR #5317, opened 2024-04-29
fix access group members resource: recreate ag members when it is del…

kavya498:
Similar to https://github.com/IBM-Cloud/terraform-provider-ibm/pull/5304


ibm/service/iamaccessgroup/resource_ibm_iam_access_group_members.go
```diff
@@ -164,6 +164,10 @@ func resourceIBMIAMAccessGroupMembersRead(context context.Context, d *schema.Res
 	listAccessGroupMembersOptions.SetLimit(limit)
 	members, detailedResponse, err := iamAccessGroupsClient.ListAccessGroupMembers(listAccessGroupMembersOptions)
 	if err != nil {
+		if detailedResponse != nil && detailedResponse.StatusCode == 404 {
```
hkantare, line 167: Does the list gives 404 when the list is empty?

kavya498: yes it gives 404 when access group itself is deleted

PR #5331, opened 2024-05-03
Refactor ibm_pi_snapshot resource

ismirlia:
This PR refactors the ibm_pi_snapshot resource and cleans up the code.

```
--- PASS: TestAccIBMPIInstanceSnapshotbasic (2037.87s)
PASS
```


ibm/service/power/ibm_pi_constants.go
```diff
@@ -33,6 +34,7 @@ const (
 	Arg_SharedProcessorPoolPlacementGroupID = "pi_shared_processor_pool_placement_group_id"
 	Arg_SharedProcessorPoolReservedCores    = "pi_shared_processor_pool_reserved_cores"
 	Arg_SnapshotID                          = "pi_snapshot_id"
+	Arg_SnapShotName                        = "pi_snap_shot_name"
```
yussufsh, line 37: In the docs, the argument is mentioned as `pi_snapshot_name`. Could you please correct it? 

Also, note in the id we use `snapshot` and in the name `snap_shot`. Hope we can correct this everywhere.

yussufsh (changes requested):

ibm/service/power/ibm_pi_constants.go
ismirlia: yea that's super awkward I'll try to fix the documentation

yussufsh (approved):
lgtm

PR #5344, opened 2024-05-10
Add Fault Attribute to Pi_Instance

michaelkad:
### Community Note

Relates OR Closes #0000

```

...
```


michaelkad:
@hkantare @yussufsh  Would please assist with this PR?
Exposing the fault attribute.

ibm/service/power/data_source_ibm_pi_instances.go
```diff
@@ -267,3 +277,20 @@ func flattenPvmInstanceNetworks(list []*models.PVMInstanceNetwork) (networks []m
 	}
 	return
 }
+
+func flattenPvmInstanceFault(fault *models.PVMInstanceFault) map[string]interface{} {
+	faultMap := make(map[string]interface{})
+	if fault.Code != 0 {
+		faultMap[Attr_Code] = strconv.FormatFloat(fault.Code, 'f', -1, 64)
+	}
```
yussufsh, line 285: Setting it directly to `Code` makes sense as the attribute value should be "0" instead of ""/null.

ibm/service/power/data_source_ibm_pi_instances.go
```diff
@@ -267,3 +277,20 @@ func flattenPvmInstanceNetworks(list []*models.PVMInstanceNetwork) (networks []m
...
+	if !fault.Created.IsZero() {
```
yussufsh, line 286: Please add a nil check.

michaelkad: There are  nil check for fault like this one, which take care of that
`if powervmdata.Fault != nil`

ibm/service/power/data_source_ibm_pi_instances.go
michaelkad: Updated

ibm/service/power/data_source_ibm_pi_instances.go
yussufsh: The fault.Created can be nil even if fault is not nil. In that case isZero will fail.

michaelkad: created is of type `strfmt.DateTime` and if omitted, it should be default to zero dateTime.

yussufsh (approved):
lgtm

PR #5420, opened 2024-06-10
Add initialize fault attribute to nil on first read

ismirlia:
Attr_Fault should be initialized to nil on first read, so that terraform apply doesn't think there's a value left to compute after vm creation.

### DIFF
--- a/ibm/service/power/resource_ibm_pi_instance.go
+++ b/ibm/service/power/resource_ibm_pi_instance.go
@@ -560,6 +560,8 @@ func resourceIBMPIInstanceRead(ctx context.Context, d *schema.ResourceData, meta
 	}
 	if powervmdata.Fault != nil {
 		d.Set(Attr_Fault, flattenPvmInstanceFault(powervmdata.Fault))
+	} else {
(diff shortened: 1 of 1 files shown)


michaelkad (approved):

ibm/service/power/resource_ibm_pi_instance.go
```diff
@@ -560,6 +560,8 @@ func resourceIBMPIInstanceRead(ctx context.Context, d *schema.ResourceData, meta
 	}
 	if powervmdata.Fault != nil {
 		d.Set(Attr_Fault, flattenPvmInstanceFault(powervmdata.Fault))
+	} else {
+		d.Set(Attr_Fault, nil)
```
hkantare, line 564: Did n't get it why we need to initialize to nil

ismirlia: If it doesn't get initialized, the terraform apply keeps thinking there's a change to apply, and tries to apply it again, but doesn't do anything. @surajibm found the issue

PR #5569, opened 2024-08-21
Update storage types docs

ismirlia:
I updated the description of the storage type/volume type/target storage tier whenever it was an argument.

Sample:
Screenshot 2024-07-12 at 2 18 35 PM


michaelkad (approved):
@hkantare, @yussufsh please help reviewing this.

PR #5669, opened 2024-09-23
[Resource] Add Volumes Bulk Attach and Detach

ismirlia:

```
--- PASS: TestAccIBMPIVolumeAttachbasic (757.61s)
PASS
--- PASS: TestAccIBMPIShareableVolumeAttachbasic (1153.82s)
PASS
```


website/docs/r/pi_volumes_attach.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+* `pi_boot_volume_id` - (Optional, String) Primary Boot Volume ID.
+* `pi_cloud_instance_id` - (Required, String) The GUID of the service instance associated with an account.
+* `pi_instance_id` - (Required, String)  The unique identifier of an instance.
+* `pi_volume_ids` - (Required, List) List of volumes to be attached to a `pi_instance`.
+
+## Attribute Reference
+
+After your resource is created, you can read values from the listed arguments and the following attributes.
+
+* `id` - The unique identifier of volumes attach. The ID is composed of `<pi_cloud_instance_id>/<pi_instance_id>/<pi_volume_ids>`
```
Alexander-Kita, line 61:
```suggestion
* `id` - (String) The unique identifier of volumes attach. The ID is composed of `<pi_cloud_instance_id>/<pi_instance_id>/<pi_volume_ids>`
```

website/docs/r/pi_volumes_detach.hmtl.markdown
```diff
@@ -0,0 +1,61 @@
...
+* `pi_detach_all_volumes` - (Optional, Boolean) Indicates if all volumes, except primary boot volume, attached to the `pi_instance` should be detached (default=`false`); required if `pi_volume_ids` is not provided.
+* `pi_detach_primary_boot_volume` - (Optional, Boolean) Indicates if primary boot volume attached to the `pi_instance` should be detached (default=`false`).
+* `pi_instance_id` - (Required, String) The unique identifier of an instance.
+* `pi_volume_ids` - (Optional, List) List of volumes to be detached from a `pi_instance`; required if `pi_detach_all_volumes` is not provided.
+
+## Attribute Reference
+
+After your resource is created, you can read values from the listed arguments and the following attributes.
+
+* `id` - The unique identifier of volumes detach. The ID is composed of `<pi_cloud_instance_id>/<pi_instance_id>`
```
Alexander-Kita, line 61:
```suggestion
* `id` - (String) The unique identifier of volumes detach. The ID is composed of `<pi_cloud_instance_id>/<pi_instance_id>`
```

website/docs/r/pi_volumes_attach.html.markdown
ismirlia: updated

website/docs/r/pi_volumes_detach.hmtl.markdown
ismirlia: updated

website/docs/r/pi_volumes_detach.html.markdown
```diff
@@ -0,0 +1,61 @@
+---
+subcategory: "Power Systems"
+layout: "ibm"
+page_title: "IBM : ibm_pi_volumes_detach"
+description: |-
+  Manages volumes in Power Virtual Server cloud.
+---
+
+# ibm_pi_volumes_attach
```
Alexander-Kita, line 9:
```suggestion
# ibm_pi_volumes_detach
```

website/docs/r/pi_volumes_detach.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+
+Detach volumes to a Power Systems Virtual Server instance, see [volumes](https://cloud.ibm.com/apidocs/power-cloud#pcloud-v2-pvminstances-volumes-post)
+
+## Example Usage
+
+The following example attach volumes to power virtual instance.
```
Alexander-Kita, line 15:
```suggestion
The following example detach volumes from a power virtual instance.
```

website/docs/r/pi_volumes_detach.html.markdown
ismirlia: updated

Alexander-Kita (dismissed):

michaelkad (dismissed):

yussufsh:
Please look at my comment on the similar feature of bulk volume delete. https://github.com/IBM-Cloud/terraform-provider-ibm/pull/5672#issuecomment-2407738202

IMO attach and detach should be part of the same resource. The volumes attached should be maintained in the state when the resource is created and removed when the resource is deleted. Having two separate resources would make things worse with errors on the next runs. Imagine you would not want to write attach and detach code in the same flow, that does not make sense OR you would not want to keep attaching or detaching only.

ismirlia:
@yussufsh 

So here's the issue we've had during development. The issue currently is that there might be a use case for detaching the primary boot volume (with the new bulk-detach call) first before attaching a new volume with new bulk-attach API. Combining the bulk-attach and bulk-detach into one resource does not let us do that.

yussufsh:
> So here's the issue we've had during development. The issue currently is that there might be a use case for detaching the primary boot volume (with the new bulk-detach call) first before attaching a new volume with new bulk-attach API. Combining the bulk-attach and bulk-detach into one resource does not let us do that.

I would suggest doing that in detach/delete func if that is required. If it is optional provide a bool param to delete the PBV.

ismirlia:
Opened: https://github.com/IBM-Cloud/terraform-provider-ibm/pull/5765 closing this one

ismirlia:
@yussufsh 

We wouldn't be able to attach a boot volume before detaching the old one if we combined it.

There could legitimately be a flow where, we create a VM, detach the automatically created boot volume, and then attach our own.

@Alexander-Kita @michaelkad

PR #5691, opened 2024-09-30
sch Agent resource updated to accommodate agent destroy resources

VaishnaviGopal:
ps: https://github.ibm.com/blueprint/product/issues/10856

### DIFF
--- a/ibm/service/schematics/resource_ibm_schematics_agent.go
+++ b/ibm/service/schematics/resource_ibm_schematics_agent.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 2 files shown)


VaishnaviGopal:
image


- Agent deploy is applied

image
image

- `run_destroy_resources` param value changed from 0-> 1 and hence when terrfaorm apply is done, a diff is found and hence it has gone to terraform update flow where `run_destroy_resources` job is performed to destroy agent deployment resources.
image


- To test Agent destroy scenario, let us deploy the agent again and try using terraform destroy

- Agent deploy is applied again this time using force parameter.

image

image
image

- Agent PRS is performed to check if cluster has agent deployment(in this case, yes. Since we did agent deploy using force with terraform.)
image

- Terraform destroy scenario: 

First agent resources should get deleted and then agent itself should get deleted
image
image
image

hkantare (approved):

PR #5723, opened 2024-10-14
bug fix

Ramya-c4:
https://github.ibm.com/devx-app-services/configuration-aggregator-planning/issues/337

### DIFF
--- a/ibm/service/configurationaggregator/data_source_ibm_config_aggregator_configurations.go
+++ b/ibm/service/configurationaggregator/data_source_ibm_config_aggregator_configurations.go
@@ -226,6 +226,5 @@ func DataSourceIbmConfigAggregatorConfigurationsConfigurationToMap(model *config
 	if err != nil {
 		return "", err
 	}
-	fmt.Println(string(jsonData))
 	return string(jsonData), nil
 }
--- a/ibm/service/configurationaggregator/data_source_ibm_config_aggregator_configurations.go
+++ b/ibm/service/configurationaggregator/data_source_ibm_config_aggregator_configurations.go
@@ -9,6 +9,7 @@ package configurationaggregator
 
 import (
 	"context"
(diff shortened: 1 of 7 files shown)


ibm/service/configurationaggregator/data_source_ibm_config_aggregator_configurations.go
```diff
@@ -77,75 +78,17 @@ func DataSourceIbmConfigAggregatorConfigurations() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"about": &schema.Schema{
-							Type:        schema.TypeList,
+							Type:        schema.TypeMap,
```
hkantare, line 81: If we go with Typemap then we have limitation from Terraform only can hold only string type. If you have combination of multiple types (string ,int, bool) then this will not be compatible.

hkantare: Can you think of declaring it string as json 
https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/resource_key#credentials_json chk this code

hkantare:
we shd have doc changes also

ibm/service/configurationaggregator/data_source_ibm_config_aggregator_configurations.go
Ramya-c4: All fields of this attribute are of type string so there is no modification  required.

Ramya-c4:
> we shd have doc changes also

The changes were made to modify the output, the input remains the same. So doc changes wouldn't be necessary.

hkantare:
It requires doc changes since we changed the schema type check personal slack for more details

hkantare (approved):

PR #5728, opened 2024-10-14
[Datasource, Resource] network address groups, NAG member.

michaelkad:
### Community Note

Relates OR Closes #0000

```
=== RUN   TestAccIBMPINetworkAddressGroupBasic
--- PASS: TestAccIBMPINetworkAddressGroupBasic (38.50s)
PASS

=== RUN   TestAccIBMPINetworkAddressGroupMemberBasic
--- PASS: TestAccIBMPINetworkAddressGroupMemberBasic (52.32s)
PASS

=== RUN   TestAccIBMPINetworkAddressGroupDataSourceBasic
--- PASS: TestAccIBMPINetworkAddressGroupDataSourceBasic (18.30s)
PASS

=== RUN   TestAccIBMPINetworkAddressGroupsDataSourceBasic
--- PASS: TestAccIBMPINetworkAddressGroupsDataSourceBasic (16.49s)
PASS
```


michaelkad:
@VaishnaviGopal , @yussufsh and @hkantare  please help with reviewing this pr.

ibm/service/power/resource_ibm_pi_network_address_group.go
```diff
@@ -0,0 +1,247 @@
...
+				Required:     true,
+				Type:         schema.TypeString,
+				ValidateFunc: validation.NoZeroValues,
+			},
+			Arg_UserTags: {
+				Description: "The user tags associated with this resource.",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Set:         schema.HashString,
+				Type:        schema.TypeSet,
```
hkantare, line 55: The IBM Terraform Provideris used in Schematics to provision templates or deployable architecture. When we provision infrastructure using Schematics it add some additional tags like workspaceid..so on.
In order to suppress the diff we need to add this custom diff

```CustomizeDiff: customdiff.Sequence(
			func(_ context.Context, diff *schema.ResourceDiff, v interface{}) error {
				return flex.ResourceTagsCustomizeDiff(diff)
			},
		),
		```
		we need to add this custom diff for all power resources where we have tags introdued

File: ibm/service/power/resource_ibm_pi_network_address_group.go [hunk 2]
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	cloudInstanceID, nagID, err := splitID(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	nagC := instance.NewIBMPINetworkAddressGroupClient(ctx, sess, cloudInstanceID)
+	networkAddressGroup, err := nagC.Get(nagID)
+	if err != nil {
+		return diag.FromErr(err)
```
hkantare, line 140: We need to handle scenario when user delete a resource outside Terraform context .
Earlier we were handling this behavior by adding Exists method but now we can handle as part of Read itself .
If get returns 404 then d.SetID to empty.

I think we need to handle this across all power resources because its missing the Exists (old behavior) or (new behaviour)
sample code
```
vpnServerRoute, response, err := sess.GetVPNServerRouteWithContext(context, getVPNServerRouteOptions)
	if err != nil {
			d.SetId("")
			return nil
		}
		log.Printf("[DEBUG] GetVPNServerRouteWithContext failed %s\n%s", err, response)
		return diag.FromErr(fmt.Errorf("[ERROR] GetVPNServerRouteWithContext failed %s\n%s", err, response))
	}
	```

Comment (yussufsh) [reply ibm/service/power/resource_ibm_pi_network_address_group.go:140]: Agreed, but when I was working with PowerVS, 404 was not implemented for all APIs, and I am sure it is still the same.
Secondly, the client calls don't return the response status code, we wrap the error messages. In Terraform we unwrap the error and find 404 in it for a few APIs already.

Comment (michaelkad) [reply ibm/service/power/resource_ibm_pi_network_address_group.go:145]: Updated

File: ibm/service/power/resource_ibm_pi_network_address_group.go (hunk 1 above)
Comment (michaelkad) [reply ibm/service/power/resource_ibm_pi_network_address_group.go:56]: @hkantare Could elaborate more on this where do we add them or give an example of this?

We are using `flex.UpdateGlobalTagsUsingCRN` which uses `Set` from what a user is provided through power, then only difference would be if the schematicsTags is the same as the one provided. At that point we can not control that part. In `flex.UpdateGlobalTagsUsingCRN` tags are appended to the list a user provided.
```
schematicTags := os.Getenv("IC_ENV_TAGS")
		var envTags []string
		if schematicTags != "" {
			envTags = strings.Split(schematicTags, ",")
			add = append(add, envTags...)
		}
```

michaelkad: NVM -- I found the example.

hkantare (approved):

PR #5730, opened 2024-10-16
Fix potential null pointer bug in pi capture user tags during creation

Alexander-Kita:
### Community Note

Relates OR Closes #0000

Description:
Fixes bug where capture image data CRN was read even though it could be nil during user tag handling in the creation function, which would cause an error.

```
=== RUN   TestAccIBMPICaptureBasic
--- PASS: TestAccIBMPICaptureBasic (81.80s)
PASS
```

```
=== RUN   TestAccIBMPICaptureUserTags
--- PASS: TestAccIBMPICaptureUserTags (124.27s)
PASS
```

### DIFF
--- a/ibm/service/power/resource_ibm_pi_capture.go
+++ b/ibm/service/power/resource_ibm_pi_capture.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	st "github.com/IBM-Cloud/power-go-client/clients/instance"
@@ -196,7 +197,10 @@ func resourceIBMPICaptureCreate(ctx context.Context, d *schema.ResourceData, met
 		imageClient := st.NewIBMPIImageClient(ctx, sess, cloudInstanceID)
 		imagedata, err := imageClient.Get(capturename)
 		if err != nil {
-			log.Printf("Error on get of ibm pi capture (%s) while applying pi_user_tags: %s", capturename, err)
+			if strings.Contains(err.Error(), NotFound) {
+				d.SetId("")
+			}
+			return diag.Errorf("Error on get of ibm pi capture (%s) while applying pi_user_tags: %s", capturename, err)
 		}
 		if imagedata.Crn != "" {
 			oldList, newList := d.GetChange(Arg_UserTags)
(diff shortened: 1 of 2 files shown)


ibm/service/power/resource_ibm_pi_capture.go
```diff
@@ -198,11 +198,13 @@ func resourceIBMPICaptureCreate(ctx context.Context, d *schema.ResourceData, met
 		if err != nil {
 			log.Printf("Error on get of ibm pi capture (%s) while applying pi_user_tags: %s", capturename, err)
 		}
-		if imagedata.Crn != "" {
-			oldList, newList := d.GetChange(Arg_UserTags)
-			err = flex.UpdateGlobalTagsUsingCRN(oldList, newList, meta, string(imagedata.Crn), "", UserTagType)
-			if err != nil {
-				log.Printf("Error on update of pi capture (%s) pi_user_tags during creation: %s", *imagedata.ImageID, err)
+		if imagedata != nil {
```
yussufsh, line 201: Don't we want to return the error in the above err condition instead of an additional check here?

Alexander-Kita: Because this is part of updating the tags on the global tagging service, we decided we do not want to throw an error since the resource has been created

yussufsh: But it concerns me that new image does not exist could mean capture did not complete successfully. 
It could be that since you are querying by name the cache on getimage api is not ready for new image. Could you check with SB why the name query fails to fetch the image details?

Alexander-Kita: I believe there is a way to check the error type but just to make sure I am understanding, you want to throw an error during create in the case of a "Not found" error, correct?

Alexander-Kita: Update: When originally testing this, the error was never hit because an image was returned correctly every time by SB at the time. I was hitting this issue today when testing it due to an error in service broker resolving the name to the incorrect ID.

Alexander-Kita: Update: When originally testing this, the error was never hit because an image was returned correctly every time by SB at the time. I was hitting this issue today when testing it due to an error in service broker. Normally this did not happen.

yussufsh: I want to ensure that the image by name works as soon as the new image is created. This condition check should not hide that fact and always work without updating the tag.

yussufsh (changes requested):

ibm/service/power/resource_ibm_pi_capture.go
```diff
@@ -198,11 +198,13 @@ func resourceIBMPICaptureCreate(ctx context.Context, d *schema.ResourceData, met
...
+			if imagedata.Crn != "" {
```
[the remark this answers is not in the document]
Alexander-Kita: Removed `if image data != nil`  error is thrown above if this is the case.

yussufsh (approved):
Will approve this.

PR #5930, opened 2025-01-21
feat(ppnlb): Support for Reserved IP as member target of PPNLB

keerthi-gp:


website/docs/r/is_lb_pool_member.html.markdown
```diff
@@ -60,6 +60,31 @@ resource "ibm_is_lb_pool_member" "example" {
...
+  target_id = ibm_is_lb.example.id
+}
+```
+
+### Sample to create a reserved ip as a member target for network load balancer.
+
+```terraform
+  resource "ibm_is_lb_pool_member" "example" {
+		lb        = ibm_is_lb.example.id
+    pool      = element(split("/", ibm_is_lb_pool.example.id), 1)
```
uibm, line 81: can you format this ?

uibm (approved):
lgtm

PR #5982, opened 2025-02-12
Deprecate network port data source and resource

ismirlia:
Screenshot 2025-02-12 at 11 58 33 AM


yussufsh (commented):
Is there any replacement or what is the reason can you mention?

ismirlia:
@yussufsh Yes they can use ibm_pi_network_interface. I have updated the message with the replacement.

yussufsh (approved):

PR #6166, opened 2025-04-17
[Datasource] Fix crash when vm has no volumes

ismirlia:
This crash occurred because the vm has no volumes and we tried to list them. I think this changed within the last year, and it is now possible to have a vm with no volumes.

This crash occurred because the vm has no volumes and we tried to list them. I think this changed within the last year, and it is now possible to have a vm with no volumes.

```
--- PASS: TestAccIBMPIVolumesDataSource_basic (28.58s)
PASS
```

List instance volumes from a vm with no volumes will look like
```
result = {
  "boot_volume_id" = tostring(null)
  "id" = "56349cd9-0269-1a8c-a609-60560b3c276d"
  "instance_volumes" = tolist([])
  "pi_cloud_instance_id" = "6021a723-bcab-4d3f-9985-d0cb2f864f35"
  "pi_instance_name" = "0ad61745-e796-4de2-bde5-89fac02f2618"
}
```


yussufsh (approved):
thanks @ismirlia

PR #6190, opened 2025-04-23
Refactor terraform errors to use new toolchain

ismirlia:
Have some questions on the new tool chain:

1. What is the benefit of the new toolchain? I'm not opposed to the new framework, I just don't understand the advantage. The issue for us at least was never, "oh we can't find where the error is." It was, "oh we misunderstood how something in terraform worked."
2. For some functions we check if the error is nil or the object returned empty. I'm assuming we have to split that in the two cases for the new toolchain cause otherwise there's no error object. I think there were a few other conditions for us where have to create the error object due to the migration.

from:
```
body := &models.SPPPlacementGroupCreate{
		Name:   &name,
		Policy: &policy,
	}

	response, err := client.Create(body)
	if err != nil || response == nil {
		return diag.Errorf("error creating the spp placement group: %v", err)
	} 
```
to:
```
body := &models.SPPPlacementGroupCreate{
		Name:   &name,
		Policy: &policy,
	}

	response, err := client.Create(body)
	if err != nil {
		tfErr := flex.TerraformErrorf(err, fmt.Sprintf("Create failed: %s", err.Error()), "ibm_pi_spp_placement_group", "create")
		log.Printf("[DEBUG]\n%s", tfErr.GetDebugMessage())
		return tfErr.GetDiag()
	}
	if response == nil {
		err = flex.FmtErrorf("response returned empty")
		tfErr := flex.TerraformErrorf(err, fmt.Sprintf("operation failed: %s", err.Error()), "ibm_pi_spp_placement_group", "create")
		log.Printf("[DEBUG]\n%s", tfErr.GetDebugMessage())
		return tfErr.GetDiag()
	}
```
?
3. Do we always replace fmt.Errorf with flex.FmtErrorF? Most of the instances of fmt.Errorf are in the terraform integration tests. Does it matter in that case?
from:
```
parts, _ := flex.IdParts(rs.Primary.ID)
		cloudpoolid := parts[0]
		placementGroupC := st.NewIBMPISPPPlacementGroupClient(context.Background(), sess, cloudpoolid)
		_, err = placementGroupC.Get(parts[1])
		if err == nil {
			return fmt.Errorf("PI SPP placement group still exists: %s", rs.Primary.ID)
		}
	}
```
to:
```
parts, _ := flex.IdParts(rs.Primary.ID)
		cloudpoolid := parts[0]
		placementGroupC := st.NewIBMPISPPPlacementGroupClient(context.Background(), sess, cloudpoolid)
		_, err = placementGroupC.Get(parts[1])
		if err == nil {
			return flex.FmtErrorF("PI SPP placement group still exists: %s", rs.Primary.ID)
		}
	}
```
4. We have some deprecated data source and resources in our provider, and there are some we know will be deprecated soon. Is it okay to skip these over?

@hkantare If you have some time please review these question.

### DIFF
--- a/ibm/service/power/data_source_ibm_pi_available_hosts.go
+++ b/ibm/service/power/data_source_ibm_pi_available_hosts.go
@@ -5,10 +5,13 @@ package power
 
 import (
 	"context"
+	"fmt"
+	"log"
 
 	"github.com/IBM-Cloud/power-go-client/clients/instance"
 
 	"github.com/IBM-Cloud/terraform-provider-ibm/ibm/conns"
+	"github.com/IBM-Cloud/terraform-provider-ibm/ibm/flex"
 	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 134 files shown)


ismirlia:
@hkantare could you please review this when you get a chance?

ismirlia:
@hkantare could you please review this when you get a chance?

hkantare:
@ismirlia Sorry for delay in response
1. What is the benefit of the new toolchain? I'm not opposed to the new framework, I just don't understand the advantage. The issue for us at least was never, "oh we can't find where the error is." It was, "oh we misunderstood how something in terraform worked."

```
This is was some new initiative started last year , new toolchain was built to standardized error messages across different platform like SDK's , Terraform. Build a centralized and standardized error codes and messages.
```
2. For some functions we check if the error is nil or the object returned empty. I'm assuming we have to split that in the two cases for the new toolchain cause otherwise there's no error object. I think there were a few other conditions for us where have to create the error object due to the migration.

```
I have seen this specific kind of behavior in power where response is nil and err is nil . Exactly not sure in which condition backend responds . May be we need to find similar approach which was stated above to handle such cases
```

3. Do we always replace fmt.Errorf with flex.FmtErrorF? Most of the instances of fmt.Errorf are in the terraform integration tests. Does it matter in that case?
```
Test case are n't mandatory to update
```

 
4. We have some deprecated data source and resources in our provider, and there are some we know will be deprecated soon. Is it okay to skip these over?
```
If we have plan to remove them in future they can be ignored
```

ismirlia:
@hkantare ok, I think have made the right changes. Please review this when you get a chance, and let me know if you have any questions.

ismirlia:
@hkantare did you get a chance to go through this?

PR #6249, opened 2025-05-19
Standardize Arg_CloudInstanceID argument across all data source and resource

ismirlia:
There were a couple of data sources that did not have the validation.NoZeroValues check that is required, and force new was added to the resources. The resources need force new on this argument because a user could accidentally update that field and think it does something when in reality it just messes up the state. Data sources don't need the value because data sources don't do updates.

```
--- PASS: TestAccIBMPIKey_basic (11.26s)
PASS
```

### DIFF
--- a/ibm/service/power/data_source_ibm_pi_datacenter.go
+++ b/ibm/service/power/data_source_ibm_pi_datacenter.go
@@ -22,9 +22,10 @@ func DataSourceIBMPIDatacenter() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			// Arguments
 			Arg_CloudInstanceID: {
-				Description: "The GUID of the service instance associated with an account.",
-				Optional:    true,
-				Type:        schema.TypeString,
+				Description:  "The GUID of the service instance associated with an account.",
+				Optional:     true,
+				Type:         schema.TypeString,
+				ValidateFunc: validation.NoZeroValues,
 			},
 			Arg_DatacenterZone: {
 				Description:  "Datacenter zone you want to retrieve.",
--- a/ibm/service/power/data_source_ibm_pi_datacenters.go
+++ b/ibm/service/power/data_source_ibm_pi_datacenters.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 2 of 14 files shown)


ibm/service/power/data_source_ibm_pi_datacenter.go
```diff
@@ -22,9 +22,10 @@ func DataSourceIBMPIDatacenter() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			// Arguments
 			Arg_CloudInstanceID: {
-				Description: "The GUID of the service instance associated with an account.",
-				Optional:    true,
-				Type:        schema.TypeString,
+				Description:  "The GUID of the service instance associated with an account.",
+				Optional:     true,
+				Type:         schema.TypeString,
+				ValidateFunc: validation.NoZeroValues,
```
yussufsh, line 28: If there is no zero value validation, then I think `Required` should have been set to true?

ismirlia: @yussufsh This endpoint is a bit weird so I went through and double checked all the cases. 

The SDK calls a slightly different endpoint depending on whether the zone targeted is off-prem or on-prem.

If the zone is off-prem (no satloc in the zone), the cloud_instance_id is not required, and doesn't do anything is set. Not required means it should be off and not set to "".

For off-prem this is valid and return the same thing:
```
data "ibm_pi_datacenter" "test" {
   pi_datacenter_zone = "dal12"
   pi_cloud_instance_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
}

data "ibm_pi_datacenter" "test" {
   pi_datacenter_zone = "dal12"
}
```

If the zone is on-prem, the cloud_instance_id is required and cannot be set to "".

So this is the only valid configuration for that case:
```
data "ibm_pi_datacenter" "test" {
   pi_datacenter_zone = "satloc_xxx_xxxxxxxxxxxxxxxxxxxx"
   pi_cloud_instance_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
}
```

And this is never a valid configuration:
```
data "ibm_pi_datacenter" "test" {
   pi_datacenter_zone = "dal12"
   pi_cloud_instance_id = ""
}

data "ibm_pi_datacenter" "test" {
    pi_datacenter_zone = "satloc_xxx_xxxxxxxxxxxxxxxxxxxx"
    pi_cloud_instance_id = ""
}
```

So that's what the code is trying to accomplish. Please let me know if you think there's a better way to go about it.

yussufsh (approved):

PR #6322, opened 2025-07-01
[Resource] [Datasource] Add Advertise and ArpBroadcast support

ismirlia:
## **This change is for the service broker release in July 21st. Please merge this for the terraform July release.**

```
--- PASS: TestAccIBMPINetworkDataSource_basic (19.23s)
PASS

--- PASS: TestAccIBMPINetworksDataSource_basic (67.66s)
PASS

--- PASS: TestAccIBMPINetworkbasic (76.25s)
PASS

--- PASS: TestAccIBMPINetworkAdvertiseArpBroadcast (108.85s)
PASS
```

Waiting on the advertise and arpbroadcast for networks to work before doing anything else.

### DIFF
--- a/website/docs/r/pi_network.html.markdown
+++ b/website/docs/r/pi_network.html.markdown
@@ -58,8 +58,8 @@ The `ibm_pi_network` provides the following [Timeouts](https://www.terraform.io/
 
 Review the argument references that you can specify for your resource.
 
-- `pi_advertise` - (Optional, String) Enable the network to be advertised.
-- `pi_arp_broadcast` - (Optional, String) Enable ARP Broadcast.
+- `pi_advertise` - (Optional, String) Enable the network to be advertised. Only supported for `vlan` network type.
+- `pi_arp_broadcast` - (Optional, String) Enable ARP Broadcast. Only supported for `vlan` network type.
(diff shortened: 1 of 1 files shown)


website/docs/r/pi_network.html.markdown
```diff
@@ -59,6 +58,8 @@ The `ibm_pi_network` provides the following [Timeouts](https://www.terraform.io/
 
 Review the argument references that you can specify for your resource.
 
+- `pi_advertise` - (Optional, String) Enable the network to be advertised.
+- `pi_arp_broadcast` - (Optional, String) Enable ARP Broadcast.
```
yussufsh: Based on `return diag.Errorf("%s, %s, and %s cannot be set when %s is pub-vlan", Arg_Advertise, Arg_ARPBroadcast, Arg_Cidr, Arg_NetworkType)` it would be good to mention that these are applicable for vlan type and should be used for pub-vlan.

yussufsh (changes requested):

website/docs/r/pi_network.html.markdown
ismirlia: Sounds good. I'll update that to make the documentation more explicit.

yussufsh (approved):

PR #6327, opened 2025-07-01
[Resource] [Datasource] Replace SSH Key API with new API 

ismirlia:
## **This change is for the service broker release in July 21st. Please merge this for the terraform release that comes after date.**

```
--- PASS: TestAccIBMPIKeysDataSourceBasic (11.38s)
PASS

--- PASS: TestAccIBMPIKeyDataSource_basic (9.00s)
PASS

--- PASS: TestAccIBMPIKey_basic (11.67s)
PASS

--- PASS: TestAccIBMPIKeyWorkspace (9.54s)
PASS

--- PASS: TestAccIBMPIKeyUpdate (21.29s)
PASS
```


ibm/service/power/resource_ibm_pi_key.go
```diff
@@ -60,14 +78,26 @@ func ResourceIBMPIKey() *schema.Resource {
 			},
 			Attr_Name: {
 				Computed:    true,
+				Deprecated:  "This field is deprecated and will be removed in a future release. Use pi_name instead.",
```
yussufsh:
```suggestion
				Deprecated:  "This field is deprecated and will be removed in a future release. Use pi_key_name instead.",
```

ibm/service/power/resource_ibm_pi_key.go
```diff
@@ -60,14 +78,26 @@ func ResourceIBMPIKey() *schema.Resource {
...
 				Description: "User defined name for the SSH key.",
 				Type:        schema.TypeString,
 			},
 			Attr_Key: {
 				Computed:    true,
+				Deprecated:  "This field is deprecated and will be removed in a future release. Use pi_key instead.",
```
yussufsh:
```suggestion
				Deprecated:  "This field is deprecated and will be removed in a future release. Use ssh_key_id instead.",
```

ismirlia: It should be "pi_ssh_key" instead of "ssh_key_id", but yes you are correct that this message is incorrect.

website/docs/d/pi_keys.html.markdown
```diff
@@ -48,5 +48,8 @@ In addition to all argument reference list, you can access the following attribu
 
   Nested scheme for `keys`:
   - `creation_date` - (String) Date of SSH Key creation.
+  - `description` - (String) Description of the ssh key.
   - `name` - (String) User defined name for the SSH key.
   - `ssh_key` - (String) SSH RSA key.
+  - `primary_workspace` - (Boolean) Indicates if the current workspace owns the ssh key or not.
```
yussufsh, line 53: Are we missing `ssh_key_id`?

ismirlia: Yes it was missing. I went ahead and added it, and tidied up some fields that were out of order. Additionally, the tests were a little outdated since the visibility default changed to "workspace" from "account", so I went ahead and updated that and re-ran all the tests.

yussufsh (approved):

PR #6505, opened 2025-10-10
Fix ibm_pi_key data source to use id

ismirlia:

```
--- PASS: TestAccIBMPIKeyDataSource_basic (11.29s)
PASS
```

### DIFF
--- a/ibm/service/power/data_source_ibm_pi_key.go
+++ b/ibm/service/power/data_source_ibm_pi_key.go
@@ -5,10 +5,12 @@ package power
 
 import (
 	"context"
+	"fmt"
+	"log"
 
 	"github.com/IBM-Cloud/power-go-client/clients/instance"
-	"github.com/IBM-Cloud/power-go-client/helpers"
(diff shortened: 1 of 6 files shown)


ibm/service/power/data_source_ibm_pi_key.go
```diff
@@ -25,11 +27,21 @@ func DataSourceIBMPIKey() *schema.Resource {
...
+				Optional:      true,
+				Type:          schema.TypeString,
+				ValidateFunc:  validation.NoZeroValues,
+			},
 			Arg_KeyName: {
-				Description:  "User defined name for the SSH key.",
-				Required:     true,
-				Type:         schema.TypeString,
-				ValidateFunc: validation.NoZeroValues,
+				AtLeastOneOf:  []string{Arg_SSHKeyID, Arg_KeyName},
```
yussufsh, line 40: Don't you want to add the deprecation message?

michaelkad: Added

PR #6506, opened 2025-10-10
Fix instance data sources to use id

ismirlia:

Deprecate the use of pi_instance_name on the ibm_pi_instance, ibm_pi_instance_console_languages, ibm_pi_instance_volumes, ibm_pi_instance_ip, and ibm_pi_pvm_snapshot data sources for pi_instance_id.

```
--- PASS: TestAccIBMPIInstanceDataSource_basic (44.64s)
PASS

--- PASS: TestAccIBMPIInstanceConsoleLanguages (18.56s)
PASS

--- PASS: TestAccIBMPIInstanceVolumesDataSource_basic (23.19s)
PASS

--- PASS: TestAccIBMPISnapshotDataSource_basic (18.25s)
PASS

--- PASS: TestAccIBMPIInstanceIPDataSource_basic (22.42s)
PASS
```

### DIFF
--- a/ibm/service/power/data_source_ibm_pi_instance_volumes.go
+++ b/ibm/service/power/data_source_ibm_pi_instance_volumes.go
@@ -35,7 +35,6 @@ func DataSourceIBMPIInstanceVolumes() *schema.Resource {
 				Description:   "The PVM instance ID.",
 				Optional:      true,
 				Type:          schema.TypeString,
-				ValidateFunc:  validation.NoZeroValues,
 			},
 			Arg_InstanceName: {
 				AtLeastOneOf:  []string{Arg_InstanceID, Arg_InstanceName},
@@ -44,7 +43,6 @@ func DataSourceIBMPIInstanceVolumes() *schema.Resource {
 				Description:   "The name of the PVM instance.",
 				Optional:      true,
 				Type:          schema.TypeString,
-				ValidateFunc:  validation.NoZeroValues,
 			},
 
 			// Attribute
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "go.mod|go.sum|.*.map|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2025-09-19T09:28:12Z",
+  "generated_at": "2025-10-02T08:57:25Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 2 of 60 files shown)


ibm/service/power/data_source_ibm_pi_instance.go
```diff
@@ -27,11 +27,22 @@ func DataSourceIBMPIInstance() *schema.Resource {
 				Type:         schema.TypeString,
 				ValidateFunc: validation.NoZeroValues,
 			},
+			Arg_InstanceID: {
+				AtLeastOneOf:  []string{Arg_InstanceID, Arg_InstanceName},
+				ConflictsWith: []string{Arg_InstanceName},
+				Description:   "The ID of the PVM instance.",
+				Optional:      true,
+				Type:          schema.TypeString,
+				ValidateFunc:  validation.NoZeroValues,
```
yussufsh: Just wondering if we should remove this NoZeroValues validation, as this is optional now.

ibm/service/power/data_source_ibm_pi_instance_volumes.go
```diff
@@ -29,11 +29,20 @@ func DataSourceIBMPIInstanceVolumes() *schema.Resource {
 				Type:         schema.TypeString,
 				ValidateFunc: validation.NoZeroValues,
 			},
+			Arg_InstanceID: {
+				AtLeastOneOf:  []string{Arg_InstanceID, Arg_InstanceName},
+				ConflictsWith: []string{Arg_InstanceName},
+				Deprecated:    "The pi_instance_name field is deprecated. Please use pi_instance_id instead",
```
yussufsh: Wrong place.

ibm/service/power/data_source_ibm_pi_instance_volumes.go
```diff
@@ -47,11 +56,26 @@ func DataSourceIBMPIInstanceVolumes() *schema.Resource {
 				Description: "List of volumes attached to instance.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						Attr_Auxiliary: {
```
yussufsh, line 59: In the future, please create a separate PR for unrelated changes.

ibm/service/power/data_source_ibm_pi_instance.go
ismirlia: fixed

ibm/service/power/data_source_ibm_pi_instance_volumes.go
ismirlia: updated

ibm/service/power/data_source_ibm_pi_instance_volumes.go
ismirlia: Sure I'll try to keep my PRs as small as possible.

PR #6508, opened 2025-10-10
Fix ibm_pi_volume data source to use id

ismirlia:

Output of acceptance testing:

```
--- PASS: TestAccIBMPIVolumeDataSource_basic (17.51s)
PASS

--- PASS: TestAccIBMPIVolumeDataSource_replication (19.34s)
PASS
```

### DIFF
--- a/ibm/acctest/acctest.go
+++ b/ibm/acctest/acctest.go
@@ -250,7 +250,7 @@ var (
 	Pi_placement_group_id             string
 	Pi_remote_id                      string
 	Pi_remote_type                    string
-	Pi_replication_volume_name        string
+	Pi_replication_volume_id          string
 	Pi_resource_group_id              string
 	Pi_route_filter_id                string
 	Pi_route_id                       string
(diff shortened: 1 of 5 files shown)


ibm/service/power/data_source_ibm_pi_volumes.go
```diff
@@ -0,0 +1,260 @@
+// Copyright IBM Corp. 2017, 2021 All Rights Reserved.
```
yussufsh: Years are older.

yussufsh: A separate PR would make sense.

ismirlia: I broke it up here: https://github.com/IBM-Cloud/terraform-provider-ibm/pull/6520

ismirlia:
@yussufsh I rebased this PR and opened another for ibm_pi_volumes data source.

hkantare (approved):

PR #6599, opened 2025-12-18
fix(instance-group): adding instance-group filter to list instances

shreyasms07:


ibm/service/vpc/data_source_ibm_is_instances_test.go
```diff
@@ -388,11 +389,15 @@ func testAccCheckIBMISInstancesDataSourceConfig1(vpcname string) string {
 		vpc_name = "%s"
 	}`, vpcname)
 }
-func testAccCheckIBMISInstancesDataSourceConfigInstanceGroup(insGrpName string) string {
+func testAccCheckIBMISInstancesDataSourceConfigInstanceGroup(insGrpName, insGrpCRN string) string {
 	return fmt.Sprintf(`
 	data "ibm_is_instances" "ds_instances1" {
 		instance_group_name = "%s"
-	}`, insGrpName)
+	}
```
uibm, line 395: This should result in duplicate data "ibm_is_instances" configuration

uibm (changes requested):

PR #6754, opened 2026-04-23
Read crash on power volumes datasource

ismirlia:
Attr_Auxiliary and Attr_ReplicationEnabled are optional to return from the swagger so they may not come back in the response. We should have nil checks before reading from them.

```
--- PASS: TestAccIBMPIInstanceDataSource_basic (28.26s)
PASS

--- PASS: TestAccIBMPIVolumesDataSourceBasic (89.45s)
PASS
```


michaelkad (approved):

uibm (approved):
lgtm

PR #6803, opened 2026-05-21
IAM Identity preferences support

briangleeson:
### Description

This PR replaces the previous PR here: https://github.com/IBM-Cloud/terraform-provider-ibm/pull/6370

- Add support for read, update, delete of Identity Preferences in IAM Identity Service.
- Fix up acceptance tests from previous PR
- Fix up examples from previous PR
- Remove unnecessary file added in the previous PR
- Also improve the effective account settings examples

### Pre-submission Checklist

- [x] Run `go mod tidy` (if you modified `go.mod`)
- [x] Run `go fmt ./...` to format all code
- [x] Run `go vet ./...` to check for common errors
- [x] All acceptance tests pass locally

```
$ make testacc TEST=./ibm/service/iamidentity TESTARGS='-run=TestAccIBMIamIdentityPreference*'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./ibm/service/iamidentity -v -run=TestAccIBMIamIdentityPreference* -timeout 700m 2>&1 | grep -v "^\[.*\]\ Set"
=== RUN   TestAccIBMIamIdentityPreferenceDataSourceBasic
--- PASS: TestAccIBMIamIdentityPreferenceDataSourceBasic (18.28s)
=== RUN   TestAccIBMIamIdentityPreferencesDataSourceBasic
--- PASS: TestAccIBMIamIdentityPreferencesDataSourceBasic (22.80s)
=== RUN   TestAccIBMIamIdentityPreferenceBasic
--- PASS: TestAccIBMIamIdentityPreferenceBasic (26.47s)
PASS
ok      github.com/IBM-Cloud/terraform-provider-ibm/ibm/service/iamidentity     77.053s
```


briangleeson:
Spotted another problem, changing this back to draft for now

Usage of `value_list_of_strings` is not working, and requires a fix in the Go SDK first

briangleeson:
The fix for the Go SDK is in this PR: https://github.com/IBM/platform-services-go-sdk/pull/448

Once that is merged and there's a new SDK version available, I can update this PR to pull in that version

briangleeson:
Found an issue with updating multiple identity_preference values in the same configuration, due to parallel requests being sent. [Adding a mutex lock](https://github.com/IBM-Cloud/terraform-provider-ibm/pull/6803/commits/5e5770f702dc4920ee6c06a6b9b2984135fff95b) for the unique ID fixed this problem

ibm/service/iamidentity/data_source_ibm_iam_account_settings.go
```diff
@@ -26,185 +26,185 @@ func DataSourceIBMIamAccountSettings() *schema.Resource {
...
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     false,
 				Description: "Defines if the entity history is included in the response.",
 			},
-			"account_id": {
+			"account_id": &schema.Schema{
 				Type:        schema.TypeString,
-				Computed:    true,
+				Required:    true,
```
hkantare: This is breaking change changing from Computed to Required , Can we mark it as optional and convert to required after few releases

ibm/service/iamidentity/resource_ibm_iam_identity_preference.go
```diff
@@ -0,0 +1,233 @@
...
+			"scope": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Scope of the preference, 'global' or 'account'.",
+			},
+		},
+	}
+}
+
+func resourceIBMIamIdentityPreferenceCreate(context context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
hkantare, line 77: This implies that we will have default preference in an account, so can we leverage this create as an update itself as we already have  "preference_id" as required input. Instead of user performing explict import this create can act as import

ibm/service/iamidentity/data_source_ibm_iam_account_settings.go
briangleeson: After discussing with my team, I'm going to revert this change as it's a breaking one

ibm/service/iamidentity/resource_ibm_iam_identity_preference.go
briangleeson: It's true that account preferences always have a default state, so cant be created as such. Do you mean to make a call to `resourceIBMIamIdentityPreferenceUpdate` here instead of returning an error? Or have some other approach in mind?

briangleeson: Actually, I'm thinking something simple like this, though it would require `iam_id` to be `Required` instead of `Computed`

```
func resourceIBMIamIdentityPreferenceCreate(context context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
	d.SetId(fmt.Sprintf("%s/%s/%s/%s",
		d.Get("account_id").(string),
		d.Get("iam_id").(string),
		d.Get("service").(string),
		d.Get("preference_id").(string),
	))
	return resourceIBMIamIdentityPreferenceRead(context, d, meta)
}
```

briangleeson: I've added these changes after testing them out. Let me know what you think @hkantare: https://github.com/IBM-Cloud/terraform-provider-ibm/pull/6803/commits/c2a411d8ef52d2f04fc5841d8a43d721a446c94a

hkantare: @briangleeson 
This approach still needs user to run apply twice (first apply -> implict import, second apply -> if user makes any changes in arguments), 
I think we call update in create , so it performs both import/updates in single apply

briangleeson: Ok @hkantare I've updated to use the update function inside create as suggested. Tested it out and seems to work as described, a single call imports and updates the resource

hkantare (approved):

PR #6849, opened 2026-06-11
Add missing Arg_SourceChecksum to image import

ismirlia:
The timeout on the create was increased to 2 days because the importing an image for the cos can take a very very long time. The test below took about 10 hours to complete, but it did actually complete. One of my coworkers from storage mentioned it could take up to several days. I put the timeout to 2 days as a result.

This is only missing on the resource.

Acceptance test output:
```
--- PASS: TestAccIBMPIImageBYOLImport (36729.31s)
PASS
```


michaelkad (approved):

ibm/service/power/resource_ibm_pi_image.go
```diff
@@ -34,7 +34,7 @@ func ResourceIBMPIImage() *schema.Resource {
 		Importer:      &schema.ResourceImporter{},
 
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(60 * time.Minute),
+			Create: schema.DefaultTimeout(2880 * time.Minute),
```
yussufsh, line 37: Did you really mean to give such a long duration? This is very high.

yussufsh (approved):

PR #6860, opened 2026-06-16
fix(iam): validate invited users

uibm:
### Community Note

Relates OR Closes #0000

### Pre-submission Checklist

- [ ] Run `go mod tidy` (if you modified `go.mod`)
- [ ] Run `go fmt ./...` to format all code
- [ ] Run `go vet ./...` to check for common errors
- [ ] All acceptance tests pass locally

```

...
```

### DIFF
--- a/ibm/service/iampolicy/resource_ibm_iam_user_invite.go
+++ b/ibm/service/iampolicy/resource_ibm_iam_user_invite.go
@@ -546,23 +546,28 @@ func resourceIBMIAMInviteUsers(d *schema.ResourceData, meta interface{}) error {
 		return InviteUserError
 	}
 
-	// poll until all invited users
-	// transitions through states: (not visible) -> PROCESSING -> PENDING or
-	// ACTIVE. ERROR_WHILE_PROCESSING means the invite workflow failed for that
-	// user; we retry in case it is transient, but fail if it persists.
-	// Valid accepted states: ACTIVE, PENDING, PROCESSING.
+	// IAM invites are async and eventually consistent. PROCESSING is transient
+	// and re-polled; PENDING means the invite was sent (success); ACTIVE means
+	// the user was already onboarded before this invite (redundant, error);
+	// ERROR_WHILE_PROCESSING means the invite failed.
 	targetEmails := make(map[string]bool, len(users))
 	for _, u := range users {
 		targetEmails[strings.ToLower(u.Email)] = true
 	}
-	var missing []string // not yet visible in account
-	var errored []string // currently in ERROR_WHILE_PROCESSING
+	var missing []string    // not yet visible in account
+	var processing []string // still PROCESSING (transient)
+	var pending []string    // PENDING — invite sent
+	var active []string     // ACTIVE — already onboarded
+	var errored []string    // ERROR_WHILE_PROCESSING — invite failed
 	scanUsers := func(allUsers []v2.UserInfo) {
 		stateByEmail := make(map[string]string, len(allUsers))
 		for _, u := range allUsers {
 			stateByEmail[strings.ToLower(u.Email)] = u.State
 		}
 		missing = missing[:0]
+		processing = processing[:0]
+		pending = pending[:0]
+		active = active[:0]
 		errored = errored[:0]
 		for email := range targetEmails {
 			state, found := stateByEmail[email]
@@ -571,13 +576,17 @@ func resourceIBMIAMInviteUsers(d *schema.ResourceData, meta interface{}) error {
 				continue
 			}
 			switch state {
-			case "ACTIVE", "PENDING", "PROCESSING":
-				// Accepted terminal (or in-progress) states — no action needed.
+			case "PENDING":
+				pending = append(pending, email)
+			case "ACTIVE":
+				active = append(active, email)
 			case "ERROR_WHILE_PROCESSING":
 				errored = append(errored, email)
+			case "PROCESSING":
+				processing = append(processing, email)
 			default:
-				// Unknown state; treat as not-yet-ready and keep retrying.
-				missing = append(missing, email)
+				// Unknown/intermediate state; treat as transient.
+				processing = append(processing, email)
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


ibm/service/iampolicy/resource_ibm_iam_user_invite.go
```diff
@@ -544,6 +545,53 @@ func resourceIBMIAMInviteUsers(d *schema.ResourceData, meta interface{}) error {
 	if InviteUserError != nil {
 		return InviteUserError
 	}
+
+	// poll until all invited users
+	// appear in the account or the timeout is reached.
```
ocofaigh: Are you only checking if they appear? I don't think this would catch the failed invite. The issue I was hitting was the user invite status was "ERROR_WHILE_PROCESSING". Valid states should be "ACTIVE", "PENDING", or "PROCESSING".

ibm/service/iampolicy/resource_ibm_iam_user_invite.go
```diff
@@ -548,46 +548,57 @@ func resourceIBMIAMInviteUsers(d *schema.ResourceData, meta interface{}) error {
 
 	// poll until all invited users
 	// appear in the account or the timeout is reached.
+	// Users with ERROR_WHILE_PROCESSING are a hard failure; stop immediately.
```
ocofaigh: I would maybe attempt the retry anyway - the key is to make sure that if its still in "ERROR_WHILE_PROCESSING" state after retry, the terraform should fail.

I would suggest to log a decent error message with as much information as possible, as its likely customer may need to open a support case if they do see "ERROR_WHILE_PROCESSING" when trying to do user invites.

ibm/service/iampolicy/resource_ibm_iam_user_invite.go
```diff
@@ -548,46 +548,57 @@ func resourceIBMIAMInviteUsers(d *schema.ResourceData, meta interface{}) error {
...
 		for _, u := range allUsers {
-			found[strings.ToLower(u.Email)] = true
+			stateByEmail[strings.ToLower(u.Email)] = u.State
 		}
 		missing = missing[:0]
+		errored = errored[:0]
 		for email := range targetEmails {
-			if !found[email] {
+			state, found := stateByEmail[email]
+			if !found || state == "PROCESSING" {
```
ocofaigh: PROCESSING is not the only valid state. "ACTIVE", "PENDING", and "PROCESSING" are all valid states.

ibm/service/iampolicy/resource_ibm_iam_user_invite.go
```diff
@@ -544,6 +545,75 @@ func resourceIBMIAMInviteUsers(d *schema.ResourceData, meta interface{}) error {
...
+		errored = errored[:0]
+		for email := range targetEmails {
+			state, found := stateByEmail[email]
+			if !found {
+				missing = append(missing, email)
+				continue
+			}
+			switch state {
+			case "ACTIVE", "PENDING", "PROCESSING":
+				// Accepted terminal (or in-progress) states — no action needed.
```
sougata-halder: When an user invited to account  its having this state transition 
PROCESSING  --> PENDING 
                         --> ERROR_WHILE_PROCESSING

Which means if an user's state in PROCESSING then it can go either on PENDING OR ERROR_WHILE_PROCESSING and its an async process so PROCESSING could stay for few secs (based on dependency call) before going either on PENDING OR ERROR_WHILE_PROCESSING

IMO,  If we see any user state is PROCESSING we should do the list call in next iteration to validate the actual stable decision making state (PENDING/ ERROR_WHILE_PROCESSING)

If an user state is ERROR_WHILE_PROCESSING then its an error state which indicate user's invitation is unsuccessful and possible process is cancel/ delete  that invite and perform re-invite 

When an user is onboarded to the account (after invite **acceptance**) then its state become ACTIVE

ibm/service/iampolicy/resource_ibm_iam_user_invite.go
```diff
@@ -544,6 +545,75 @@ func resourceIBMIAMInviteUsers(d *schema.ResourceData, meta interface{}) error {
...
+	}
+	retryErr := resource.Retry(5*time.Minute, func() *resource.RetryError {
+		allUsers, listErr := client.ListUsers(accountID)
+		if listErr != nil {
+			return resource.RetryableError(listErr)
+		}
+		scanUsers(allUsers)
+		if len(errored) > 0 {
+			// ERROR_WHILE_PROCESSING may be transient — keep retrying.
+			return resource.RetryableError(fmt.Errorf("user(s) %v are in ERROR_WHILE_PROCESSING state", errored))
```
sougata-halder: With this are you doing re invite those users by cancelling the existing invite ?

sougata-halder (commented):
Left comments 

Usually if an invite is PENDING then it success 
If an invite is ERROR_WHILE_PROCESSING its an invite failure

ibm/service/iampolicy/resource_ibm_iam_user_invite.go
sougata-halder: @uibm  Can you please help with this comment

sougata-halder (commented):
Overall looks good , Can you help me with the comment 
https://github.com/IBM-Cloud/terraform-provider-ibm/pull/6860/changes#r3568150907

ibm/service/iampolicy/resource_ibm_iam_user_invite.go
uibm: no, it is not re-inviting or cancelling anything.

uibm: It just re-reads the account's user list and re-checks the state. On ERROR_WHILE_PROCESSING it returns a RetryableError, which tells the SDK "call the loop function again in a bit" but the next iteration only lists again. There's no RemoveUsers (cancel/delete) and no InviteUsers (re-invite) happening.

uibm: added a commit, which removes the polling status completely

uibm:
this has been fixed as part of #6902

PR #6861, opened 2026-06-16
Enable DHCP Private

michaelkad:
### Community Note

Relates OR Closes #0000

### Pre-submission Checklist

- [x] Run `go mod tidy` (if you modified `go.mod`)
- [x] Run `go fmt ./...` to format all code
- [x] Run `go vet ./...` to check for common errors
- [x] All acceptance tests pass locally

```
=== RUN   TestAccIBMPINetworkEnableDHCPSatellite
--- PASS: TestAccIBMPINetworkEnableDHCPSatellite (126.76s)
PASS

=== RUN   TestAccIBMPINetworkDataSource_basic
--- PASS: TestAccIBMPINetworkDataSource_basic (50.85s)
PASS
```


website/docs/d/pi_networks.html.markdown
```diff
@@ -50,7 +50,8 @@ In addition to all argument reference list, you can access the following attribu
   - `advertise` - (String) Indicates if the network is advertised.
   - `arp_broadcast` - (String) Indicates if ARP Broadcast is enabled.
   - `crn` - (String) The CRN of this resource.
-  - `dhcp_managed` - (Boolean) Indicates if the network DHCP Managed.
+  - `dhcp_managed` - (Boolean) DHCP enabled network.
```
yussufsh, line 53: Any reason this is missing in `pi_network`?

website/docs/d/pi_networks.html.markdown
```diff
@@ -50,7 +50,8 @@ In addition to all argument reference list, you can access the following attribu
...
+  - `enable_dhcp` - (Boolean) DHCP enabled network.
```
yussufsh, line 54: More precise information would be helpful, which is supported on on-prem locations.

website/docs/d/pi_networks.html.markdown
michaelkad: @yussufsh  To match the `NetworkCreate` struct, as we will deprecate it later on. Since they resolve to the same thing in the backend, we are not including it.

website/docs/d/pi_networks.html.markdown
michaelkad: This is a common feature for on-prem and off-prem

yussufsh: But if I am the user of this resource what should I refer to? If both are same then we may need to deprecate one.

yussufsh: I said on-prem because it is mentioned in the argument section.

michaelkad: > But if I am the user of this resource what should I refer to? If both are same then we may need to deprecate one.

@yussufsh  I got this answer "yes keep both for now. We have not discussed the deprecation timeline with OCP team yet"  

> I said on-prem because it is mentioned in the argument section.

You’re right; initially, this was only available on-prem, which is why it was mentioned. I’ve updated the pr to remove it. That should have been removed in the first place, an oversight on my part.

yussufsh: np, thanks.

yussufsh (approved):

PR #6936, opened 2026-08-04
fix(vpc): treat unhealthy as a pending state when waiting for instance group health

### LINKED ISSUE #6935: ibm_is_instance_group: waitForHealthyInstanceGroup treats 'unhealthy' as terminal, failing applies the API already accepted
### Description

`waitForHealthyInstanceGroup` in `ibm/service/vpc/resource_ibm_is_instance_group.go` declares:

```go
healthStateConf := &resource.StateChangeConf{
    Pending: []string{SCALING},
    Target:  []string{HEALTHY},
    ...
}
```

An instance group reports `unhealthy` while a member is being added, replaced, or is still booting. Since `unhealthy` is in neither `Pending` nor `Target`, `StateChangeConf` classifies it as an unexpected state and returns as soon as it observes one. Terraform then fails the apply even though the API call succeeded and the group converges to `healthy` shortly afterwards.

This is not the timeout expiring, so raising `timeouts` does not help — the watcher gives up on first sight of the state rather than at a deadline. We have seen the same class of change abort after ~40s on one occasion and after ~5m10s on another.

### Impact

`waitForHealthyInstanceGroup` is shared, so this affects:

- `ibm_is_instance_group` — create (`:286`), update (`:405`) and the pre-delete wait (`:574`)
- `ibm_is_instance_group_manager`
- `ibm_is_instance_group_manager_policy`
- `ibm_is_instance_group_manager_action`

Any change that adds or cycles members can trip it: `instance_count`, `subnets`, or an `instance_template` swap. The practical effect is a red pipeline on a change that actually succeeded, requiring a manual re-run.

### Reproduction

1. Create an `ibm_is_instance_group` whose members take more than a few seconds to boot and become healthy.
2. Change `instance_count` (or `subnets`, or the `instance_template`).
3. `terraform apply` fails with:

```
Error: waitForHealthyInstanceGroup failed: unexpected state 'unhealthy',
wanted target 'healthy'. last error: %!s(<nil>)
```

4. The change has nonetheless been applied. The group reaches `healthy` on its own a short time later — in our case about two minutes after Terraform gave up — and re-running `apply` with an unchanged configuration succeeds.

### Expected behaviour

`unhealthy` should be a **pending** state: keep polling until the group is `healthy` or the configured timeout expires. That is what the `timeouts` block is for, and it matches how transient states are handled elsewhere in the VPC resources. A group that genuinely cannot become healthy would then fail at the timeout instead of immediately — a slower failure, but a correct one, and tunable per configuration.

### Versions

- Provider: 1.89.0
- Also present on `master` as of this writing.

I have a small patch for this and will open a PR shortly.

cchristous:
### Community Note

Closes #6935

### What

`waitForHealthyInstanceGroup` listed only `scaling` as a pending state, so `StateChangeConf` classified `unhealthy` as an unexpected state and returned the moment it observed one. An instance group passes through `unhealthy` whenever a member is added, replaced, or is still booting — so an `instance_count`, `subnets` or `instance_template` change on a group whose members are slow to boot fails the apply, after the API has already accepted the change and shortly before the group converges to `healthy` on its own.

This adds an `UNHEALTHY` constant and includes it in `Pending`.

### Why this is the right fix

The wait's contract is "block until the group is healthy or the timeout expires". `unhealthy` is a state the group legitimately passes through on the way to `healthy`, so it belongs in `Pending` rather than being treated as terminal.

Raising `timeouts` is not a substitute, because the abort is not a deadline expiring — the wait exits at whatever moment it first observes the state. We measured aborts at ~40s and at ~5m10s for the same class of change, and in one case the group reached `healthy` about two minutes after Terraform had already given up.

`deleting` is deliberately **not** added to `Pending`: a group being deleted while we wait for it to become healthy will never get there, and should keep failing fast.

### Behaviour change

A group that can never become healthy now fails when the timeout expires rather than immediately. That is a slower failure, but it is the semantics the `timeouts` block already advertises, and callers who want a fast failure can set a short `create`/`update` timeout.

### Scope

`waitForHealthyInstanceGroup` is shared, so this also covers `ibm_is_instance_group_manager`, `ibm_is_instance_group_manager_policy` and `ibm_is_instance_group_manager_action`, plus the create, update and pre-delete waits of `ibm_is_instance_group`.

### Pre-submission Checklist

- [x] Run `go mod tidy` (if you modified `go.mod`) — not applicable, `go.mod` is unchanged
- [x] Run `go fmt ./...` to format all code — `gofmt -l` reports no files needing formatting
- [x] Run `go vet ./...` to check for common errors — exits 0 with no findings
- [ ] All acceptance tests pass locally

I have **not** run the instance-group acceptance tests. They provision real VSIs in a live IBM Cloud account, which I am not set up to do for this resource, and I would rather say so than imply coverage I do not have. The change is a two-line adjustment to a state machine's pending list, with no new API calls, no schema change and no change to the request payload. I am happy to run them if a maintainer can point me at the expected `TF_ACC` configuration for `ibm_is_instance_group`, or to defer to CI.

### DIFF
--- a/ibm/service/vpc/resource_ibm_is_instance_group.go
+++ b/ibm/service/vpc/resource_ibm_is_instance_group.go
@@ -26,6 +26,8 @@ const (
 	SCALING = "scaling"
 	// HEALTHY ...
 	HEALTHY = "healthy"
+	// UNHEALTHY ...
+	UNHEALTHY = "unhealthy"
 	// DELETING ...
 	DELETING                     = "deleting"
 	isInstanceGroupAccessTags    = "access_tags"
@@ -660,7 +662,12 @@ func waitForHealthyInstanceGroup(instanceGroupID string, meta interface{}, timeo
 	getInstanceGroupOptions := vpcv1.GetInstanceGroupOptions{ID: &instanceGroupID}
 
 	healthStateConf := &resource.StateChangeConf{
-		Pending: []string{SCALING},
+		// An instance group reports "unhealthy" while a member is being added, replaced, or is
+		// still booting, so it is a state on the way to "healthy" rather than a terminal one.
+		// Leaving it out of Pending makes StateChangeConf return the moment it observes one,
+		// failing an apply the API has already accepted. "deleting" stays out deliberately: a
+		// group being deleted while we wait for it to become healthy will never get there.
+		Pending: []string{SCALING, UNHEALTHY},
 		Target:  []string{HEALTHY},
 		Refresh: func() (interface{}, string, error) {
 			instanceGroup, response, err := sess.GetInstanceGroup(&getInstanceGroupOptions)
--- a/ibm/service/vpc/resource_ibm_is_instance_group.go
+++ b/ibm/service/vpc/resource_ibm_is_instance_group.go
@@ -662,11 +662,13 @@ func waitForHealthyInstanceGroup(instanceGroupID string, meta interface{}, timeo
 	getInstanceGroupOptions := vpcv1.GetInstanceGroupOptions{ID: &instanceGroupID}
 
 	healthStateConf := &resource.StateChangeConf{
-		// An instance group reports "unhealthy" while a member is being added, replaced, or is
-		// still booting, so it is a state on the way to "healthy" rather than a terminal one.
-		// Leaving it out of Pending makes StateChangeConf return the moment it observes one,
-		// failing an apply the API has already accepted. "deleting" stays out deliberately: a
-		// group being deleted while we wait for it to become healthy will never get there.
+		// "unhealthy" means the group has fewer than membership_count instances, which is also
+		// true while members are being added, replaced, or are still booting, so it is a state on
+		// the way to "healthy" rather than a terminal one. Leaving it out of Pending makes
+		// StateChangeConf return the moment it observes one, failing an apply the API has already
+		// accepted. A group the platform itself considers broken is caught by lifecycle_state in
+		// Refresh instead. "deleting" stays out deliberately: a group being deleted while we wait
+		// for it to become healthy will never get there.
 		Pending: []string{SCALING, UNHEALTHY},
 		Target:  []string{HEALTHY},
 		Refresh: func() (interface{}, string, error) {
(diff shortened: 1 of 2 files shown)


ibm/service/vpc/resource_ibm_is_instance_group.go
```diff
@@ -660,7 +662,12 @@ func waitForHealthyInstanceGroup(instanceGroupID string, meta interface{}, timeo
 	getInstanceGroupOptions := vpcv1.GetInstanceGroupOptions{ID: &instanceGroupID}
 
 	healthStateConf := &resource.StateChangeConf{
-		Pending: []string{SCALING},
+		// An instance group reports "unhealthy" while a member is being added, replaced, or is
```
deepakibms: Instance group could get into unhealthy state other than while members are getting added or replaced, in which case the resource will wait until timeout. A recommendation is,  LifecycleState value LifecycleStateReasons can be used to verify the state along with the status value check.

cchristous: Thanks — good catch, and adopted in fdc03315: the refresh now returns immediately when `lifecycle_state` is `failed` or `suspended`, reporting the `lifecycle_reasons` that explain it. That matches how `ibm_is_instance_network_attachment`, `ibm_is_dedicated_host` and `ibm_is_placement_group` already treat those two states.

One part I couldn't settle from the API definitions, and would value your read on: I don't think `lifecycle_state` can distinguish a *transient* membership shortfall from a persistent one. `status = unhealthy` is defined as "group is unable to reach `membership_count` instances", and the documented `lifecycle_reasons` codes are `internal_error` and `resource_suspended_by_provider` — both platform-level rather than member-level. So a group whose member cannot be placed (say, no capacity for the profile in that zone) looks like it stays `lifecycle_state = stable` with `status = unhealthy`. If that's right, gating the pending state on `lifecycle_state` would either keep failing applies the API has accepted, or not change the outcome — which is why I've used it only to fail fast.

That does leave the case you raised: a group that stays unhealthy for a reason the API doesn't surface will now consume the timeout rather than returning quickly. I think that's the right trade — it's the wait-then-report contract the rest of the provider follows, and it's what makes the common case correct. On our workload the group reliably reaches `healthy` a couple of minutes *after* the current code has already failed the apply (measured: replacement member created ~2 min after `terraform apply` aborted, group `healthy` at full membership shortly after).

So the question back to you: does the service ever report `lifecycle_state != stable` when the problem is member placement? If it does, I'll gate the pending state on that instead of keeping `unhealthy` pending unconditionally.

deepakibms: Yes lifecycle_state does not distinguish a transient membership based error from a persistent one.

cchristous: Thanks for confirming — that settles it, so I've left the fail-fast as the only use of `lifecycle_state` and kept `unhealthy` pending unconditionally.

Two things I can't do from this side when you get a chance:

1. **Approve the workflow run.** The current head (`3d1d8e4`) has no checks at all — only the first commit ever got a `Build` run (it passed). It looks like the fork-contributor approval gate, so CI can't report until a maintainer releases it.
2. **An approving review**, if you're happy with the change as it stands.

For completeness on testing: the new unit test covers the lifecycle-reason formatting, but the wait behaviour itself isn't reachable from an acceptance test — it needs a group that actually goes `unhealthy`, which can't be provisioned on demand. The evidence for it is a production measurement, in the PR description.

deepakibms (approved):

ibm/service/vpc/resource_ibm_is_instance_group.go
deepakibms: One another point that popped in my mind, just to be sure, is the Delete flow which touches upon the same-for-healthy method which may need some testing on a lifecycle_state="failed" group specifically with the update on membership_count to zero. This should be good as well but just to want to verify once. I am checking with the service team as well. If updating membership_count could make the lifecycle_state to transition back to heathy and if API allows deleting a failed group directly, then delete path need not go through the wait-for-healthy method.

uibm (approved):

PR #7013, opened 2026-09-21
chore: "fix for sms and email subscriptions to manage invitee via invited parameter"

Divya-Singh1693:
Email Subscription create and  update
sub-2
sub-3
sub1
sub-5
sub-7
sub-9
sub-11
sub-4
sub-6
sub-8


ibm/service/eventnotification/resource_ibm_en_subscription_sms.go
```diff
@@ -226,6 +241,12 @@ func resourceIBMEnSMSSubscriptionRead(context context.Context, d *schema.Resourc
 		return diag.FromErr(fmt.Errorf("[ERROR] Error setting updated_at: %s", err))
 	}
 
+	if result.Attributes != nil {
```
uibm, line 244: attributes are getting set here and is optional in the schema. Can you cross check this ?

ibm/service/eventnotification/resource_ibm_en_subscription_sms.go
```diff
@@ -62,21 +62,36 @@ func ResourceIBMEnSMSSubscription() *schema.Resource {
...
+							Computed:    true,
+							Description: "Phone numbers that have unsubscribed. Populated by the service; read-only.",
 							Elem:        &schema.Schema{Type: schema.TypeString},
 						},
+						"add": {
+							Type:       schema.TypeList,
+							Optional:   true,
+							Computed:   true,
+							Deprecated: "Use invited to manage phone numbers.",
+							Elem:       &schema.Schema{Type: schema.TypeString},
```
uibm, line 86: its marked deprecated but the code has completely removed it, what will happen to the existing users ?

uibm (changes requested):

ibm/service/eventnotification/resource_ibm_en_subscription_sms.go
Divya-Singh1693: This is not part of recent change although logically attribute is required for subscription. Is this gonna impact anyway in current change?

ibm/service/eventnotification/resource_ibm_en_subscription_sms.go
Divya-Singh1693: Any change made to these will not be applied, but discarded silently. But user can manage these via invited hence it will still provide them way to make changes.

uibm: if user earlier had add and remove

they still have this in their config

now in the read you are not setting it as empty

so this should keep on showing change on every plan ?

ibm/service/eventnotification/resource_ibm_en_subscription_sms.go
uibm: earlier attributes was optional

now you are setting it in read meaning optional and computed

This can be both okay/not okay depending on what user has in his config ?

Divya-Singh1693: Updated attributes to be Required, since it is required for user to provide subscription inputs in order to create subscription.

ibm/service/eventnotification/resource_ibm_en_subscription_sms.go
Divya-Singh1693: I have updated the code to ignore provided entries in resource block for add and remove parameters. This will help them maintain state if no entries are added, but it will also not reflect in any change on cloud resource even if entries are added under add and remove.

